PR #14, opened 2020-02-27
config: modify the host IP and port to env vars

hihilary:


Judge40:
How will we handle default values for local development? Will we have to remember to add to the .env files?

hihilary:
@Judge40 fixed! Thanks a lot. Babul told me to just keep one variables ADDRESS. Also modified it.

PR #22, opened 2020-04-09
Feature: Form-R part-a

babul3012:


hihilary:
There're some conflicts with master due to some other merges. Please merge from master first.

src/__snapshots__/App.snapshot.spec.tsx.snap
```diff
@@ -10,7 +10,7 @@ Array [
       className="nhsuk-width-container nhsuk-header__container"
     >
       <div
-        className="nhsuk-header__logo"
+        className="nhsuk-header__logo nhsuk-header__logo--only"
```
hihilary, line 13: May I ask what this snapshot file used for, please?

babul3012: This is part of snapshot testing - https://jestjs.io/docs/en/snapshot-testing
These snapshots help us to make sure that UI is not changed until we wanted it to.

babul3012:
> There're some conflicts with master due to some other merges. Please merge from master first.

Done.

src/services/TraineeProfileService.ts
```diff
@@ -0,0 +1,16 @@
+import { AxiosResponse } from "axios";
+import ApiService from "./apiService";
+import { TraineeProfile } from "../models/TraineeProfile";
+
+export class TraineeProfileService {
+  private apiService: ApiService;
+  constructor() {
+    super("/trainee/api");
```
hihilary, line 8: This TraineeProfileService doesn't have a parent class, but calls super() in constructor. And SonarCloud reports a bug here as well.

babul3012: Updated.

PR #25, opened 2020-04-15
fix: Updated services to use context paths in url

babul3012:
TISNEW-4247

### DIFF
--- a/src/services/apiService.ts
+++ b/src/services/apiService.ts
@@ -5,9 +5,9 @@ export class ApiService {
   axiosInstance: AxiosInstance;
   traineeTisId = 123;
 
-  constructor() {
+  constructor(baseUrl: string) {
     this.axiosInstance = axios.create({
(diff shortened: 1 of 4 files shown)


src/services/apiService.ts
```diff
@@ -3,11 +3,12 @@ import { Auth } from "aws-amplify";
 
 export class ApiService {
   axiosInstance: AxiosInstance;
+  urlPath: string = "";
   traineeTisId = 123;
 
-  constructor() {
+  constructor(urlPath: string) {
+    this.urlPath = urlPath;
     this.axiosInstance = axios.create({
```
Judge40, line 11: Why not set the baseUrl here?

babul3012: I thought baseUrl needs to be absolute. Relative path is getting added at the trail of the base url.
Will modify it.

PR #31, opened 2020-04-22
refactor: Added common validation schema

babul3012:


src/components/forms/formr-part-a/ValidationSchema.tsx
```diff
@@ -6,120 +6,75 @@ const mobileRegex = /((\+44(\s\(0\)\s|\s0\s|\s)?)|0)7\d{3}(\s)?\d{6}/g;
...
+    "Anticipated completion date"
+  ).test(
+    "completionDate",
+    "Anticipated completion date - please choose a future date",
+    value => DateUtilities.IsFutureDate(value)
+  ),
+  trainingGrade: stringSchema("Training Grade"),
+  startDate: dateSchema("Programme start date"),
+  programmeMembershipType: stringSchema("Post type / Appointment"),
   wholeTimeEquivalent: yup
```
john12321, line 74: Still not sure about the validation for the WTE. For example, the error we get for whitespace is not very user-friendly. A cleaner solution might be a single regex to test the string input - something like this: `/^((0\.[1-9]{1})?|(0\.([0-9]{1}[1-9]{1}|[1-9]{1}[0-9]{1}))|1(\.0{1,2})?)$/` - with just a single error message. Happy to push suggested changes to this branch if agreed. Cheers

babul3012: @john12321 Sure. Please push the change.

PR #40, opened 2020-05-07
Validation and label name fixes

babul3012:
* Updated warning message text
* Enforcing selection when program / local office is not found in reference data


StevenHee:
Spotted a wonderful typo in the warning message text - "By singing this document..."

PR #44, opened 2020-05-15
feat: Added inline error messaging for text and select input components

StevenHee:
TISNEW-4531

- Modified the existing components **TextInputField** and **SelectInputField** to include components from the NHS React library that support inline error messaging and hint text. https://nhsdigital.github.io/nhsuk-react-components/?path=/story/input--with-error-string

- Added option to render multi-line html textarea instead of input field by setting a rows property   

- Updated error summary in line with NHS React component to improve accessibility through improved labelling.

- Added new component, **InputFooterLabel** to allow insertion of additional text below the form field that supports html markup to allow for inclusion of links.

- Added missing label properties to the address form fields on formr part a.

- Added hint text to email form field on formr part b, advising on preference for NHS.net email addresses.  

- Updated version of nhsuk-react-components in package.json. **This package update is required in order to render the new components correctly.**

### DIFF
--- a/src/components/forms/TextInputField.tsx
+++ b/src/components/forms/TextInputField.tsx
@@ -15,11 +15,12 @@ interface Props {
 }
 
 const TextInputField: FunctionComponent<Props> = props => {
-  const [field, { error, touched }] = useField({
-    name: props.name,
-    type: props.name
-  });
+  const [field, { error, touched }] = useField(props);
   const FormElement = props.rows ? Textarea : Input;
+
+  const setFieldWidth = (width: number) => {
+    return width < 20 ? 20 : Math.floor(width / 10) * 10;
+  };
 
   return (
     <>
@@ -31,10 +32,14 @@ const TextInputField: FunctionComponent<Props> = props => {
         }
       >
         <FormElement
-          width={props.width || 20}
+          width={
+            field.value ? props.width || setFieldWidth(field.value.length) : 20
+          }
           error={error && touched ? error : ""}
           id={props.id || props.name}
-          {...field}
+          onBlur={field.onBlur}
+          onChange={field.onChange}
+          value={field.value}
           {...props}
         />
         <InputFooterLabel label={props.footer || ""} />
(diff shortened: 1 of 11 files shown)


babul3012:
> Good job mate!
> Please check and fix the tests.

nvm.. I updated it. Snapshot had to be updated as we have updated the nhs library which changed the logo code a bit.

src/components/forms/formr-part-b/Sections/Section1.tsx
```diff
@@ -35,6 +35,7 @@ const Section1 = (props: any) => {
             <TextInputField
               label="Primary contact email address"
               name="email"
+              hint="For reasons of security and due to frequent systme failures with internet email accounts, you are strongly advised to provide an NHS.net email address."
```
john12321, line 38: Typo: _frequent **systme** failures_

src/components/forms/TextInputField.tsx
```diff
@@ -1,14 +1,46 @@
...
+  hint?: string;
+  width?: any;
+  footer?: any;
+  type?: string;
+}
+
+const TextInputField: FunctionComponent<Props> = props => {
+  const [field, { error, touched }] = useField({
+    name: props.name,
+    type: props.name
```
john12321, line 20: Not sure what this does.

StevenHee:
Cheers guys. I've been looking into the input field issue that Johno has flagged.

I agree there is an issue with text inputs where the field is pre-populated with data and the length exceeds the width of the input such that the text is hidden. For fields where we might expect the text to exceed the default width, we could simply set the width property on that input. Alternatively, if no width property is set, we could set the width of the field based on the length of the text. My preference is to avoid setting all inputs to full width, which is also inline with recommendations.

The select input appears to work as expected and auto-sizes depending on the width of the longest option label. The issue arises on smaller screens such as mobile whereby the text is truncated. This appears to be unavoidable and is a limitation of the select input type since the text doesn't wrap. To prevent this issue we could either shorten the option label or implement using radio buttons where labels can wrap.

src/components/forms/formr-part-b/Sections/Section1.tsx
babul3012: I will fix this in my commit.

PR #104, opened 2020-06-08
Add more e2e tests to Form R PartB

john12321:
- Add e2e tests for form R section 3
- Add more tests to section 2
- Some refactoring of existing tests
- Fix some bugs with existing tests now failing intermittently. One issue was the view not defaulting to mobile view when required. This issue was fixed by removing this command from the the cypress index file and running it separately in each test file that required a mobile view.  

Ticket: TISNEW-4315

### DIFF
--- a/cypress/integration/formr-b/formr-b.spec.ts
+++ b/cypress/integration/formr-b/formr-b.spec.ts
@@ -48,7 +48,6 @@ describe("Form R (Part B)", () => {
       .should("not.be.empty");
     // intentionally left email blank
     cy.get("#email").should("be.visible");
-    cy.wait(1000);
     cy.get("#localOfficeName > option")
       .eq(1)
       .then(element => {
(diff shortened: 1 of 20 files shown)


cypress/integration/formr-b/formr-b.spec.ts
```diff
@@ -119,13 +133,115 @@ describe("Form R (Part B)", () => {
...
+    cy.get(
+      ".nhsuk-pagination-item--next > .nhsuk-pagination__link > .nhsuk-pagination__page"
+    ).click();
+    cy.get("[data-cy=gmcNumber]").should("have.text", "22222222");
+
+    cy.get(':nth-child(2) > [type="submit"]').should("be.visible").click();
+
+    // See list of saved forms
+    cy.contains("Submitted forms").should("be.visible");
+    cy.get("[data-cy=submittedForm0]").should("be.visible");
```
babul3012, line 245: May be hypothetical based on our approach for tests - what if we have some saved forms already? Doesn't this check for the first row which might not be the current saved form?

john12321: Yeah, thinking was based on a testing env where only one saved form would exist. But in the meantime it will, as you say, check to see if there are any saved forms (there should be at least one) by getting the first in the list.

PR #146, opened 2020-07-20
test: Updated formr-b save functionality test

babul3012:
* Added case for logout and login

### DIFF
--- a/cypress/integration/formr-b/formr-b.spec.ts
+++ b/cypress/integration/formr-b/formr-b.spec.ts
@@ -278,11 +278,14 @@ describe("Form R (Part B)", () => {
         cy.checkAndFillSection1(currRevalDate, prevRevalDate);
         cy.get("[data-cy=BtnSaveDraft]").click();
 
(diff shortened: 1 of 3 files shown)


cypress/integration/formr-b/formr-b.spec.ts
```diff
@@ -278,6 +278,11 @@ describe("Form R (Part B)", () => {
         cy.checkAndFillSection1(currRevalDate, prevRevalDate);
         cy.get("[data-cy=BtnSaveDraft]").click();
 
+        cy.logout();
+        cy.login();
+        cy.get("[data-cy=BtnMenu]").should("be.visible").click();
+        cy.contains("Form R-b").click();
+
```
heetis-shivanirana, line 285: @babul3012  How are we verifying here that the changes we made before log out still exist after we login again in the application?

babul3012: Added condition to check the gmcNumber. Pushed the changes.

PR #194, opened 2020-08-26
Add cache-busting

john12321:
**Problem** 
The app is registering a service-worker which means that there are issues with browser caching. For example, the service worker cache will only be cleared if the app is opened in a new tab. If not then the user will keep the old version of the app.

**Solution**
Add cache-busting code to: 
- generate a meta.json file on each build that contains the latest version number - and which isn't cached by the browser 
- compare latest version on the meta.json file with the cached client version
- unregister Service Worker, clear cache, reload page if there is a version mismatch.

TISNEW-5121


StevenHee (approved):

generate-build-version.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const appVersion = packageJson.version;
+
+const jsonData = {
+  version: appVersion
+};
+
+var jsonContent = JSON.stringify(jsonData);
+
+fs.writeFile("./public/meta.json", jsonContent, "utf8", (err: any) =>
```
paulhoang-hee, line 13: does the build fail if an error / console.log(err) happen?

PR #215, opened 2020-09-21
refactor: Migrate api service to new API endpoints

Judge40:
A new set of API endpoints has been created with a shared `/api` parent
instead of individual `/api` children. Migrate to use the new endpoints,
which also no longer need the trainee's TIS ID to be provided.

TISNEW-3833

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -477,7 +477,7 @@ Cypress.Commands.add("checkFlags", name => {
   return cy
     .request({
       method: "GET",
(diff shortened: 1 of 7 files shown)


Judge40:
These changes will cause some local disruption as the endpoints no longer map directly to what is used in the service. The nginx config in the dev handbook will need to be updated to handle it similar to how we redirect local requests for TIS endpoints.

babul3012 (approved):

john12321 (approved):

PR #223, opened 2020-09-23
fix: Type in trainee profile error message

Judge40:
The error message displayed when no trainee data is available in the
trainee profile is incorrectly spelt.

NO-TICKET


john12321 (approved):
congito!

PR #324, opened 2021-02-05
Test/add e2e pipeline

StevenHee:
Add e2e task with mocha reporting. Outputs created as artifacts and uploaded to S3 artifacts bucket. Slack notification sent to notifications-e2e-test-results channel.

Ticket: An addendum to https://hee-tis.atlassian.net/browse/TIS21-1020

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 src/serviceWorker.ts
 build/
+reports/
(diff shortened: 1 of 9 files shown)


Judge40 (changes requested):
Looks really good, but we should change `deploy-prod` to depend on `cypress-e2e`

StevenHee:
> Looks really good, but we should change `deploy-prod` to depend on `cypress-e2e`

Cheers Andy. I just wondered if we wanted to prevent E2E from being a blocker to deploying to prod to avoid similar issues encountered with admins-ui.

Judge40 (approved):

PR #341, opened 2021-03-05
fix: so e2e task shows as failed in github

StevenHee:
The E2E task in the deploy action included a continue-on-error flag so fails were not showing in GitHub. This flag has been removed and additional conditionals have been added to make sure the artifacts are still generated and backed-up and Slack message sent. The Slack messaging has been separated for success and failures and styled accordingly.  

NO TICKET

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -81,10 +81,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Get workflow
+        run: |
+          echo "workflow=$(echo ${{ github.workflow }} | sed -e 's/\W/-/g' -e 's/\(.*\)/\L\1/')" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -81,10 +81,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Get workflow
```
Judge40, line 84: Need a blank line here.

Judge40 (approved):

Judge40 (changes requested):
Approved by accident, see previous comment.

.github/workflows/deploy.yml
StevenHee: > Need a blank line here.

Cheers Andy. Line break added.

Judge40 (approved):

PR #353, opened 2021-03-14
fix:update container to defined tag

StevenHee:
E2E tests were failing due to running against a version of Chrome that was greater than 3 versions behind the current, resulting in the banner being displayed and concealing some inputs. The container `cypress/browser:latest` appears not to be latest having been updated 5 months ago and runs Chrome 83 so I have set this as a the actual latest container tag. Going forward, I don't think we should rewrite the tests to accommodate the presence of the banner since we don't support that browser version, but we will need to update the container tag periodically as the version of chrome is updated, or find an alternative approach.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -77,7 +77,7 @@ jobs:
     needs: [deploy-pre-prod]
     name: Run Cypress E2E tests
     runs-on: ubuntu-latest
-    container: cypress/browsers:latest
+    container: cypress/browsers:node14.16.0-chrome89-ff86
 
     steps:
       - uses: actions/checkout@v2


Judge40:
Ouch, thats rough. Dependabot can do actions versions but I doubt it can handle this, feels awkward having our tests periodically start failing.

Could we disable the browser check on stage instead?

StevenHee:
> Ouch, thats rough. Dependabot can do actions versions but I doubt it can handle this, feels awkward having our tests periodically start failing.
> 
> Could we disable the browser check on stage instead?

We could disable the feature on stage. However, if we're defining what versions of browsers we support shouldn't we make sure we are running E2E against these?  Alternatively, we could define a chrome release in the browser check that matches the container version such that the banner never shows.

Judge40:
> > Ouch, thats rough. Dependabot can do actions versions but I doubt it can handle this, feels awkward having our tests periodically start failing.
> > 
> > Could we disable the browser check on stage instead?
> 
> We could disable the feature on stage. However, if we're defining what versions of browsers we support shouldn't we make sure we are running E2E against these?  Alternatively, we could define a chrome release in the browser check that matches the container version such that the banner never shows.    

I agree we should be using a supported browser for tests, thats what using the `latest` tag is expected to give us I guess.
Do we know how far behind that image is? I'd rather avoid having a specific version defined if we can.

Judge40 (approved):
Looks like dependabot will eventually support docker containers in the GHA workflow file, so I'm happy to merge this and deal with the manual updates in the short term.

PR #362, opened 2021-03-17
fix: error when non medical curriculum

Judge40:
When the trainee's latest programme had no medical curriculums the
FormA/B failed to load, due to a TypeError on the `reduce()` function
called on the empty array.

Perform a sort instead of reduce to get the latest curriculum as it
better handles the empty array.

TIS21-1327

### DIFF
--- a/src/models/ProfileToFormRPartAInitialValues.ts
+++ b/src/models/ProfileToFormRPartAInitialValues.ts
@@ -22,9 +22,12 @@ export function ProfileToFormRPartAInitialValues(
     programme && programme.curricula.length > 0
       ? programme.curricula
           .filter(c => c.curriculumSubType === MEDICAL_CURRICULUM)
-          .reduce(function (a, b) {
-            return a.curriculumStartDate > b.curriculumStartDate ? a : b;
-          })
+          .sort(
+            (a, b) =>
+              new Date(b.curriculumStartDate).getTime() -
+              new Date(a.curriculumStartDate).getTime()
+          )
+          .shift()
       : null;
 
   const model: FormRPartA = {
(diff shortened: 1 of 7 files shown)


src/models/ProfileToFormRPartAInitialValues.ts
```diff
@@ -22,9 +22,12 @@ export function ProfileToFormRPartAInitialValues(
     programme && programme.curricula.length > 0
       ? programme.curricula
           .filter(c => c.curriculumSubType === MEDICAL_CURRICULUM)
-          .reduce(function (a, b) {
-            return a.curriculumStartDate > b.curriculumStartDate ? a : b;
-          })
+          .sort(
+            (a, b) =>
+              new Date(b.curriculumStartDate).getTime() -
+              new Date(a.curriculumStartDate).getTime()
```
Judge40, line 28: Unsure why I needed to initialize a new date here, but `curriculum.curriculumStartDate.getTime()` errored.

src/models/ProfileToFormRPartBInitialValues.ts
```diff
@@ -22,9 +22,12 @@ export function ProfileToFormRPartBInitialValues(
     programme && programme.curricula.length > 0
       ? programme.curricula
           .filter(c => c.curriculumSubType === MEDICAL_CURRICULUM)
-          .reduce(function (a, b) {
-            return a.curriculumStartDate > b.curriculumStartDate ? a : b;
-          })
+          .sort(
+            (a, b) =>
+              new Date(b.curriculumStartDate).getTime() -
+              new Date(a.curriculumStartDate).getTime()
```
Judge40, line 28: Unsure why I needed to initialize a new date here, but `curriculum.curriculumStartDate.getTime()` errored.

src/models/ProfileToFormRPartAInitialValues.ts
john12321: It's JS magic.
Or it's because initialising a **new Date()** object gives us access to the getTime method while **Date()** gives us a string and so only those methods available to strings.
JS magic sounds better though.

src/models/ProfileToFormRPartBInitialValues.ts
john12321: more magic

john12321 (approved):

PR #370, opened 2021-03-26
Fix/use reference dbc for form rb previous body

Judge40:


Judge40:
Converted to draft to check/update E2E tests

john12321 (approved):

PR #394, opened 2021-04-12
fix: Add pre-prop specialty tie breaker to FormRs

Judge40:
The pre-populated specialty for Form R Part A/B is based on the lastest
programme's latest curricula. In the case of a tie, i.e. two curricula
starting on the same day the specialty pre-populated is random.
Apply an alphabetical sort as a tie breaker, the first specialty
alphabetically should be the on populated.

TIS21-1397


john12321 (approved):
Looks fine @Judge40 . The only thing that we might want to look at next (follow-up ticket?) is to refactor both intialValues files to keep them DRYer/ keep sonarcloud happier. 
e.g. put the methods used to assign variables programme, curriculum etc. in a utils file? Happy to look at this as it was FEers who created this duplication in the first place 😄

PR #397, opened 2021-04-16
feat: Replace form switches with feature flags

Judge40:
Feature flags have been implemented in the backend as a cleaner
alternative to the form switches approach. Replace the form switches
implementation with feature flags.

TIS21-1452


Judge40:
Changes are ready to go, but reliant on https://github.com/Health-Education-England/tis-trainee-forms/pull/195 to provide the endpoint with data before it can be merged.

I wasn't sure the best practice for if the feature flag endpoint failed, currently the value is set to `null` which is later treated as false.

StevenHee:
Has the covid property stored in mongo db changed?

Judge40:
> Has the covid property stored in mongo db changed?

This technique would supersede that, with the form switches property being removed completely once this is merged

StevenHee:
OK. So where is the COVID property being set?

Judge40:
> OK. So where is the COVID property being set?

It's a property in the forms service, can be set by an env var when starting the service. Defaults to `true` if that env var is not present.
https://github.com/Health-Education-England/tis-trainee-forms/pull/195/files#diff-7bc26be01d3e7c9c566f2a28dc1b5cd87dbdd5cb619184f46eb2bf223bf7825dR24

StevenHee:
So does changing this property require releasing a new version of the service?

Judge40:
Essentially, yes.
While we can toggle it via the env var we have to deploy a new task definition and restart the service to do that. Locally we can just change the env var and restart the service.

Even as a DB entry, the value should only be managed by the forms service anyway rather than manually being modified directly in the DB (no change tracking of behaviour being modified if we change it manually).

The history from last week:
We had the issue that the COVID declaration flag was never set programmatically, it was a manual insert to the DB which isn't ideal and led to the covid declaration not appearing on production (because no one ever added it) 

After creating a way to programmatically insert/update the DB entry it was thought that having the flags be DB backed didn't really offer anything, it was just another layer of complication, where as using config backed flags keeps it much simpler.

StevenHee:
Thanks Andy. That's really helpful.

john12321 (approved):
Just tested in local dev env - works great.  👍 @Judge40

PR #401, opened 2021-04-22
fix: abbreviate contact names for mobile view

john12321:
- Make support contact drop-down names readable on mobile view.

TIS21-1494


john12321:
It's an epic PR - might need a comfy chair and a coffee!

StevenHee (approved):

PR #410, opened 2021-04-26
chore: Tag the docker image with `latest`

Judge40:
The build docker image should be tagged with `latest` as well as the
commit sha.

TIS21-1460


Judge40:
This is needed to allow easier updating of the dev environment via docker compose.

Liban-H (approved):

PR #455, opened 2021-05-18
Revert "chore(deps): bump yup from 0.29.3 to 0.32.9"

ReubenRobertsHEE:
Reverts Health-Education-England/tis-trainee-ui#390


john12321 (approved):
@ReubenRobertsHEE Jumped back over from admins-ui so happy to carry the dependabot baton for a bit to give you a break. Will have a look at yup breaking changes unless you've already done this? Cheers!

ReubenRobertsHEE:
Thanks @john12321 - no, I haven't looked into it yet, so if you have a chance that would be great!

PR #459, opened 2021-05-27
Fix/dependabot prettier

StevenHee:
Following prettier update via dependabot, the build was failing in the dockerize GHA step. To fix this, I have run Prettier CLI locally which has slightly reformatted several files. This enabled the build to run locally so have deployed the reformatted files.

### DIFF
--- a/src/components/forms/formr-part-b/View.tsx
+++ b/src/components/forms/formr-part-b/View.tsx
@@ -39,13 +39,8 @@ const mapStateToProps = (state: RootState) => ({
 
 class View extends React.PureComponent<ViewProps> {
   render() {
-    const {
-      formData,
-      history,
-      editSection,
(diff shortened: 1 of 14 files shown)


src/components/forms/formr-part-b/View.tsx
```diff
@@ -39,13 +39,8 @@ const mapStateToProps = (state: RootState) => ({
...
 class View extends React.PureComponent<ViewProps> {
   render() {
-    const {
-      formData,
-      history,
-      editSection,
-      canEdit,
-      featureFlags
-    } = this.props;
+    const { formData, history, editSection, canEdit, featureFlags } =
```
john12321, line 42: Glad prettier sorted this out!

john12321 (approved):

PR #515, opened 2021-08-06
Add filtering for the list of subspecialties in the form r part a declarations section tis21 1913

EdwardBarclay:


Marcellofabbri:
Although it works, I'm not sure that adding a type-specific property (`curriculumSubType`) to an interface (`KeyValue`) is sensible, as it's meant to be the same for each of the props.

E.g.: `immigrationStatus` (as well as all of the other props that are of type `KeyValue`) also having `curriculumSubType` as a property

Marcellofabbri:
An alternative way of doing it that comes to mind would be to have something like `allCurricula` and `medicalCurricula` as Props, and then in the action that makes the api call (`getCurricula()`) make two dispatches so that one is responsible for furthering the whole payload that'll fill `allCurricula` and the other can take the payload, filter it by `curriculumSubType`, and further it filtered to fill `medicalCurricula`.
This way you'd have `allCurricula` and `medicalCurricula` among the Props, still in KeyValue form, and correctly separated. One dropdown can then have `medicalCurricula` as its `options` while the other two can have `allCurricula`.

It should work but I don't know how good it sounds as far as good practice is concerned.

PR #519, opened 2021-08-10
Tis21 1913 add filtering for the list of subspecialties in the form r part a declarations section

EdwardBarclay:


Marcellofabbri (commented):
It does seem to work fine. I've just left one comment on the CurriculumKeyValue interface possibly inheriting from KeyValue.

PR #542, opened 2021-09-09
Fix/validation formr a

StevenHee:
Update form validation onChange/onBlur  on Form R Part A to only fire after submission. This prevents the form initially loading with validation errors flagged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.33.0",
+  "version": "0.33.1",
   "private": true,
   "dependencies": {
     "@aws-amplify/ui-components": "^1.7.3",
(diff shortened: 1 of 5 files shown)


Judge40 (approved):

StevenHee:
Whilst fixing the form a errors appearing on open, I discovered that the autocomplete component was allowing users to enter values not present in the list of options. This has been fixed and a flag also added that allows for custom values if required.  

The  errors were appearing due to the onInputChange event firing when a pre-populated value is set triggering the Formik  validateOnChange.  No change event is passed on this initial firing which is used as the condition on this event firing. If allowCustomInput is false, the input value is set to empty if it does not match an item in the list.

john12321:
Nice work, @StevenHee. Will be interesting to see which validation method trainees prefer: validation on submit (Part A); or validation on change (Part B). Literal A/B testing in action!

john12321 (approved):

PR #557, opened 2021-09-27
feat: Update support info and QR code reload logic

john12321:
Quick fix to:
- Add conditional WarningCallout message for 3-minute QR code expiry.
- Redo additional support messaging.

NO TICKET


Judge40 (approved):
LGTM

PR #610, opened 2022-01-10
Feat/set totp issuer

StevenHee:
- Add suitable issuer title for TSS TOTP account in authenticator apps.
- In Authy, each account includes an icon, which is set based on Google Image Search, apparently. When the issuer title is set to Trainee Self-Service, the icon chosen is currently a stressed zebra! To remedy this, I've added 'NHS' at start which changes the the icon to something less inappropriate.

### DIFF
--- a/src/components/authentication/mfa/SetupTOTP.tsx
+++ b/src/components/authentication/mfa/SetupTOTP.tsx
@@ -31,12 +31,10 @@ const SetupTOTP = ({ user, mfaStatus, incrementStep }: ISetupMFA) => {
     const getUser = async () => {
       try {
         const totpCode: string = await Auth.setupTOTP(user);
-        const authCode: string =
-          "otpauth://totp/AWSCognito:" +
-          user.getUsername() +
-          "?secret=" +
-          totpCode +
-          "&issuer=AWSCognito";
+        const name = "NHS Trainee Self-Service";
+        const authCode: string = `otpauth://totp/${encodeURI(
+          name
+        )}:${user.getUsername()}?secret=${totpCode}`;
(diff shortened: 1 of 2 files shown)


Judge40 (changes requested):
I think it's recommended to still include the `issuer` parameter.

Info from Google Auth: https://github.com/google/google-authenticator/wiki/Key-Uri-Format#issuer

src/components/authentication/mfa/SetupTOTP.tsx
```diff
@@ -31,12 +31,10 @@ const SetupTOTP = ({ user, mfaStatus, incrementStep }: ISetupMFA) => {
     const getUser = async () => {
       try {
         const totpCode: string = await Auth.setupTOTP(user);
-        const authCode: string =
-          "otpauth://totp/AWSCognito:" +
-          user.getUsername() +
-          "?secret=" +
-          totpCode +
-          "&issuer=AWSCognito";
+        const name = "NHS Trainee Self-Service";
```
JosephKelly, line 34: If it's only used in one place, is there any value in defining it as a a `const`?  The convention in Java would be to inline the value.

StevenHee: Fair point. I was using twice, as issuer query string parameter as well. However, since this doesn't appear necessary I shall pass inline.

StevenHee:
Retained name as variable and also added as issuer querystring param.

Judge40 (approved):

JosephKelly (approved):
Resolved the convo for best practice comment

PR #665, opened 2022-05-26
Remove validation from supervisor name and email address in covid declaration

EdwardBarclay:
Remove validation from supervisor name and email address in Covid Declaration

TIS21-3067


john12321 (approved):
Looks good @EdwardBarclay. One thing to think about are the 2 new code smells. Could maybe make a utils method for when there's nothing recorded?  
Anyway, given it's an urgent fix, we could do some refactoring in a follow-up ticket perhaps? ps. That whole view section needs refactoring!

PR #682, opened 2022-06-29
fix: Add mapbox-gl to deps

john12321:
There is a known aws-amplify issue which looks to have caused the latest build to fail (https://github.com/aws-amplify/amplify-ui/issues/1736 for more info).
The fix will be in the next release: @aws-amplify/ui-react 3.0 .
Current workaround is to add **mapbox-gl** dep directly to package.json and cross your toes.

TIS21-3193

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.48.0",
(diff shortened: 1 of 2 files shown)


EdwardBarclay (approved):
LGTM

PR #680, opened 2022-06-29
Feat/rework totp

john12321:
- Separate the TOTP user journey into 2 main steps: install the Authenticator App (no time pressure); add the 'NHS TIS Self-Service' account to the Authenticator App and verify/log in (3 min before QR code expires) 
- Rework routing - add 'MFA setup' navigation item to e.g. allow for more easy updating of MFA type
- Incorporate MFA logic into redux architecture (e.g. easier to debug) - which includes success/error notification msgs
- Increase code coverage (approx. 10%) by adding new MFA tests  

TODO
- Follow-up tech improvement ticket(s) to update some libraries to latest major versions (e.g. Cypress, react-router-dom, nhsuk-react-components)
- Update the support website with new TOTP screenshots
- Refactor View comp (long-standing code smell !) and possibly others (e.g. MFA section navigation/submission)

PR #683, opened 2022-07-04
Remove phone_number attribute when swapping MFA type from SMS to TOTP

john12321:
- Remove phone_number attribute when swapping MFA type from SMS to TOTP.
- Order the actions so the phone_number removal is the last action in TOTP set-up process just in case process is aborted.
- Refactor some mfa comps

TODO 
- (see https://github.com/Health-Education-England/tis-trainee-ui/pull/680 )
- refactor redux slices to make DRY
- look at ways to increase test coverage for sign-up/MFA 
 
TICKET TIS21-3193

### DIFF
--- a/cypress/integration/mfa/mfa.spec.ts
+++ b/cypress/integration/mfa/mfa.spec.ts
@@ -1,0 +1,44 @@
+/// <reference types="cypress" />
+/// <reference path="../../support/index.d.ts" />
+
+describe("MFA set-up", () => {
+  before(() => {
+    cy.wait(30000);
+    cy.visit("./profile");
+    cy.signIn();
+  });
+  it("should render the Choose MFA page", () => {
+    cy.get('[data-cy="MFA set-up"]').should("exist").click();
+    cy.get("[data-cy=mfaAlreadyWarning]").should("exist");
+    cy.get("[data-cy=mfaAlreadyText]")
+      .should("exist")
+      .should(
+        "include.text",
+        "You have already set up your Authenticator App for MFA "
+      );
+    cy.get("[data-cy=mfaSummary]").click();
+    cy.get("[data-cy=mfaText] > :nth-child(3) > b").should(
+      "include.text",
+      "or SMS"
+    );
+    cy.get("[data-cy=mfaChoice1]").click();
+    cy.get("[data-cy=BtnSubmitMfaChoice]").click();
+    cy.get("[data-cy=backLink]")
+      .should("exist")
+      .should("contain.text", "Start over")
+      .click();
+    cy.get("[data-cy=mfaChoice0]").should("exist").click();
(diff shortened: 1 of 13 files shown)


ReubenRobertsHEE:
For what it's worth, it all looks ok to me 😁 . Just wondering, though: is it possible to actually include a test that the phone number has been removed when switching from SMS to TOTP? I'm thinking something like possibly an end-to-end(?) test of 'set up user with SMS authentication and phone number, trigger change from SMS to TOTP, retrieve user details from cognito, see that there is now no phone number'. I don't know if this is excessive in relation to the other tests, or if I've missed something where it is already tested!

john12321:
> For what it's worth, it all looks ok to me 😁 . Just wondering, though: is it possible to actually include a test that the phone number has been removed when switching from SMS to TOTP? I'm thinking something like possibly an end-to-end(?) test of 'set up user with SMS authentication and phone number, trigger change from SMS to TOTP, retrieve user details from cognito, see that there is now no phone number'. I don't know if this is excessive in relation to the other tests, or if I've missed something where it is already tested!

Hi @ReubenRobertsHEE , cheers for the feedback! Yeah this is the main gap in the MFA testing. We haven't got/can't set up the e2e infra for the TOTP journey even before getting to the point of checking the user pool phone_number attribute. For example,  for an e2e test, I think we would need a way of reading the newly-generated QR code to add 'HEE TIS Self-Service' account, key in the 6-digit code from that linked account and then do the checking in the user pool. Not sure how to do this. Any ideas gratefully received! 😄

Judge40 (approved):

Judge40:
> > For what it's worth, it all looks ok to me 😁 . Just wondering, though: is it possible to actually include a test that the phone number has been removed when switching from SMS to TOTP? I'm thinking something like possibly an end-to-end(?) test of 'set up user with SMS authentication and phone number, trigger change from SMS to TOTP, retrieve user details from cognito, see that there is now no phone number'. I don't know if this is excessive in relation to the other tests, or if I've missed something where it is already tested!
> 
> Hi @ReubenRobertsHEE , cheers for the feedback! Yeah this is the main gap in the MFA testing. We haven't got/can't set up the e2e infra for the TOTP journey even before getting to the point of checking the user pool phone_number attribute. For example, for an e2e test, I think we would need a way of reading the newly-generated QR code to add 'HEE TIS Self-Service' account, key in the 6-digit code from that linked account and then do the checking in the user pool. Not sure how to do this. Any ideas gratefully received! 😄

It would be great to be able to do automate it, I have no idea how we would start though. The effort vs reward doesn't feel great with what we know so far.

ReubenRobertsHEE (approved):
Good to have a wider tech discussion around how to get the best out of the testing frameworks for this, as we chatted about 👍

PR #686, opened 2022-07-12
feat: Update support to better direct users

john12321:
- Update support section to better direct users to appropriate place
(Currently lots of 'unsubmit form' queries going to tech support)

NO TICKET 
(just an exasperated PO)


EdwardBarclay (approved):
LGTM

PR #690, opened 2022-07-15
Update dockerfile node

john12321:
- node v12 end-of-life so upped build version to v14

NO TICKET


johnsimmons1980 (approved):
looks good from my house

PR #703, opened 2022-08-25
Feat: only include next future placement in form R part B TIS21-3361

ReubenRobertsHEE:
On FormR PartB. It is possible for there to be multiple placements on
the same day (though this is an edge case for bad data).

TIS21-3361: Restrict the placements pre-populated on Form R Part B


john12321 (approved):
Looks good, nice work, @ReubenRobertsHEE .

PR #707, opened 2022-09-06
Add date submitted stamp to both forms to be reflected in the PDFs

EdwardBarclay:
TIS21-3448

### DIFF
--- a/src/components/forms/formr-part-b/View.tsx
+++ b/src/components/forms/formr-part-b/View.tsx
@@ -865,7 +865,7 @@ const View = ({ canEdit, history }: IView) => {
             </Card>
             <h3>
               Form Submitted on:&nbsp;
-              {DateUtilities.ToUTCDate(formData.submissionDate)}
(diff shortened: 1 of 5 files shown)


src/components/forms/formr-part-b/View.tsx
```diff
@@ -864,6 +863,10 @@ const View = ({ canEdit, history }: IView) => {
                 </SummaryList>
               </Card.Content>
             </Card>
+            <h3>
+              Form Submitted on:&nbsp;
+              {DateUtilities.ToUTCDate(formData.submissionDate)}
```
john12321, line 868: Looks good, @EdwardBarclay ! 
Do you think it's worth adding a small test for the submission date visibility? (e.g. formr-part-a/View.cy-test.tsx ) 

Just had a look at the ticket...I know you have followed the format given (UTC).  Just wondering if it would be better to keep things consistent and use local date format? Just a thought.

john12321 (approved):
LGTM @EdwardBarclay

PR #709, opened 2022-09-12
refactor: Form R Part B View Section 2-8

DorisWongNHS:
TIS21-3069


src/components/forms/formr-part-b/viewSections/ViewSection1.tsx
```diff
@@ -89,7 +84,7 @@ const ViewSection1 = ({ makeSectionEditButton, formData }: IViewSection1) => {
               <SummaryList.Key>Dual Specialty</SummaryList.Key>
               <SummaryList.Value>{formData.dualSpecialty}</SummaryList.Value>
             </SummaryList.Row>
-          </SummaryList>{" "}
+          </SummaryList>
```
ReubenRobertsHEE, line 87: I wonder what that space was in there for? It makes me think it might be some browser compatibility weirdness. Might be worth checking the history and/or mentioning to Johno that we'd like to remove it?

DorisWongNHS: Double confirmed with @john12321 . The space was being used between the two <SummaryList>, now the duplicated summaryList is removed, so the space can also be removed.
https://github.com/Health-Education-England/tis-trainee-ui/blob/main/src/components/forms/formr-part-b/View.tsx#L151

john12321 (approved):
LGTM @DorisWongNHS 👍

PR #710, opened 2022-09-13
Fix: refactor form r part b view

ReubenRobertsHEE:
The approach we agreed was to refactor based on the form sections, and to add some basic e2e tests.


src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
```diff
@@ -0,0 +1,87 @@
...
+      cy.get(".nhsuk-summary-list__value").should(
+        "include.text",
+        formattedBoolean
+      );
+    } else {
+      cy.get(".nhsuk-summary-list__value").should("include.text", dataValue);
+    }
+  }
+}
+
```
ReubenRobertsHEE, line 44: 😵‍💫 not a fan of the below, but I might need some help fixing this

ReubenRobertsHEE:
🎉 it builds! Still a bit of cleaning-up to do I think

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { submittedFormRPartBs } from "../../../../mock-data/submitted-formr-partb";
+import store from "../../../../redux/store/store";
+import ViewSection1 from "./ViewSection1";
+import history from "../../../navigation/history";
+import ViewSectionShouldIncludeThisData, {
+  ISectionDataField
+} from "./ViewSectionTestHelper";
+
+const makeSectionEditButton = (section: number) => {
+  return false;
```
ReubenRobertsHEE, line 13: @john12321 this was intended as a bit of a placeholder function, but it did mean I had to change the signature of the IViewSection[x] from returning `false | JSX.Element` to  `boolean | JSX.Element`.
Not sure if (a) this is appropriate and (b) we might need tests for this

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const formDataToDisplay: ISectionDataField[] = [
+  { fieldName: "forename", format: "" },
+  { fieldName: "surname", format: "" },
+  { fieldName: "gmcNumber", format: "" },
+  { fieldName: "email", format: "" },
+  { fieldName: "localOfficeName", format: "" },
+  { fieldName: "currRevalDate", format: "LocalDate" },
```
ReubenRobertsHEE, line 22: I tried to get the property types directly from the FormRPartB model interface, but I couldn't get it to work, hence the rather awkward repetition of the 'format' attribute. There might be a better way of tackling this.

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
ReubenRobertsHEE: The issue I ran into, is that I can't reference `formData[formDataItem.fieldName]` if it is possible that `formDataItem.fieldName` is not a property of formData. Even with the `if (formDataItem.fieldName in formData) {` check, the IDE was not happy. I think there's some subtlety around the loss of typing in the conversion to 'raw' javascript that I don't really understand.

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
```diff
@@ -0,0 +1,112 @@
+import { mount } from "@cypress/react";
```
ReubenRobertsHEE, line 1: Ugly, since its not obviously used, but otherwise all the references to `cy` threw up errors.

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
```diff
@@ -0,0 +1,112 @@
...
+  fieldName: FormRPartBDeclarationField;
+  format: string;
+}
+
+export interface ISectionCovidDeclarationDataField {
+  fieldName: FormRPartBCovidDeclarationField;
+  format: string;
+}
+
+function CheckDataIsDisplayed(dataValue: any, format: string) {
```
ReubenRobertsHEE, line 36: Unsure of the benefits of defining this as` const CheckDataIsDisplayed = (dataValue: any, format: string) => {` (or something else more idiomatic for React!)

Also: naming conventions for this file and its contents?

src/mock-data/submitted-formr-partb.ts
```diff
@@ -228,3 +228,91 @@ export const submittedFormRPartBsWithUnsubmitted = [
...
+        declarationType: "Complaint",
+        dateOfEntry: "2020-06-12",
+        title: "Current declaration title",
+        locationOfEntry: "Current declaration location of entry"
+      }
+    ],
+    currentDeclarationSummary: "",
+    compliments: "",
+    haveCovidDeclarations: true,
+    covidDeclarationDto: {
```
ReubenRobertsHEE, line 301: This content has been tailored to suit the test, which expects all the fields to be included in the output. As such, the test for Section 7 does not test any of the internal display logic of the component. I'm not sure if we should look at creating a few different test data records for this that allow us to validate the display logic, or if this is a bit OTT at this point...

src/components/forms/formr-part-b/Create.cy-test.tsx
```diff
@@ -82,7 +82,7 @@ describe("Create form B", () => {
       "Section 1: Doctor's details"
     );
     cy.get(
-      ":nth-child(4) > .nhsuk-card__content > :nth-child(2) > :nth-child(1) > .nhsuk-summary-list__value"
+      ":nth-child(3) > .nhsuk-card > .nhsuk-card__content > .nhsuk-summary-list > :nth-child(1) > .nhsuk-summary-list__value"
```
ReubenRobertsHEE, line 85: Doris identified that this test was actually referencing the View component; so fixed here :)

src/components/forms/formr-part-b/viewSections/View.tsx
```diff
@@ -0,0 +1,88 @@
...
+        {enableCovidDeclaration ? <ViewSection7 {...viewSectionProps} /> : null}
+        {!canEdit && <ViewSection8 {...viewSectionProps} />}
+      </>
+    );
+  else content = <Redirect to="/formr-b" />;
+
+  return <div>{content}</div>;
+};
+
+export default View;
```
ReubenRobertsHEE, line 88: I guess we should write some tests for this! 😄

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
john12321: Hi @ReubenRobertsHEE , just pulled the branch to have a look. Re above, makes sense changing to boolean type. Will have a closer look at this and tests etc.. 
One obvious bit of refactoring I noticed in this file (and other files) is to define the `formData` and `viewSectionProps` vars within the describe block and also use  `beforeEach` inside describe to mount `ViewSection1` .

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
john12321: @ReubenRobertsHEE I think adding this to the top of the file instead should do the trick: 
`// load type definitions that come with Cypress module`
`/// <reference types="cypress" />`

src/components/forms/formr-part-b/Create.cy-test.tsx
john12321: Nice work @DorisWongNHS !

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
john12321: @ReubenRobertsHEE, only advantage I've found with using `function` syntax is when you've got a file with your functions chaotically ordered and JS saves your bacon by hoisting them! 🤣 As far a I remember there's also some difference in the way `this` is scoped too. 
As far as naming conventions go, I wonder if we can make those `ViewSection...ShouldIncludeThisData` functions into one generic function to test our TS skills to the limit! 😅

src/mock-data/submitted-formr-partb.ts
john12321: @ReubenRobertsHEE Sounds like a good idea - depends how much energy we've got after tacking the other stuff. Might be a one for another day? 😄

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
john12321: @ReubenRobertsHEE no quick answer for this one! 🤣

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
DorisWongNHS: Yeah it doesn't look good. Learning from others code from the project, I find `/// <reference types="cypress" />` at the top of the file would help this.

ReubenRobertsHEE: Nice - thanks @john12321 and @DorisWongNHS

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
ReubenRobertsHEE: That would be super-cool, the repetition is a bit 💩 . I might need a bit of help on that (or time to go a-digging, but that might be any amount of time!)

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
ReubenRobertsHEE: Done, thanks @john12321

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
john12321: Re subset/mapped types, wondering if a combo of Pick and Partial is worth exploring? Or we could always hide everything in a pure cypress/support/commands.js file? Lol

src/mock-data/submitted-formr-partb.ts
DorisWongNHS: Have just added some tests for the displaying logic in Section 7.

src/components/forms/formr-part-b/viewSections/View.tsx
DorisWongNHS: Tests added for the View page.

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
```diff
@@ -0,0 +1,89 @@
...
+import history from "../../../navigation/history";
+import ViewSectionShouldIncludeThisData, {
+  ISectionSomeDataField
+} from "./ViewSectionTestHelper";
+
+const makeSectionEditButton = (section: number) => {
+  return false;
+};
+
+const formDataToDisplay: ISectionSomeDataField[] = [
```
ReubenRobertsHEE, line 16: This now uses a single interface for all useable fields. It reads simpler, but it does remove a 'guardrail' against including a field that is not actually part of the applicable set of fields, e.g. including a field from the Work sub-component in a test that examines the main FormRPartB component.

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
ReubenRobertsHEE: 'Resolved' by using a type of `:any` for the formData object. Some other refactoring has been done, which is a bit of an improvement, but might still be a way to go.

src/components/forms/formr-part-b/viewSections/ViewSectionTestHelper.tsx
```diff
@@ -0,0 +1,75 @@
...
+  fieldName:
+    | FormRPartBKeys
+    | FormRPartBWorkKeys
+    | FormRPartBDeclarationKeys
+    | FormRPartBCovidDeclarationKeys;
+  format: string;
+}
+
+export default function CheckDataIsDisplayed(
+  formDataToDisplay: ISectionSomeDataField[],
```
ReubenRobertsHEE, line 47: Johno suggested something like 
`export function genericThing<T>(formDataToDisplay: T[], formData:any){formDataToDisplay.forEach((formDataItem: any) => {…`
Might be worth exploring!

src/components/forms/formr-part-b/viewSections/ViewSection1.cy-test.tsx
ReubenRobertsHEE: now reverted somewhat

PR #719, opened 2022-09-28
refactor: create submission date utility and update references

EdwardBarclay:
refactor the submission date methods into a FormR Utility and update tests

No-Ticket

### DIFF
--- a/src/components/forms/formr-part-a/View.cy-test.tsx
+++ b/src/components/forms/formr-part-a/View.cy-test.tsx
@@ -53,13 +53,13 @@ describe("View", () => {
     cy.get("[data-cy=BtnEdit]").should("not.exist");
     cy.get("[data-cy=BtnSaveDraft]").should("not.exist");
     cy.get("[data-cy=BtnSubmit]").should("not.exist");
-    cy.get("#__cy_root > :nth-child(1) > :nth-child(3)").should("exist");
-    cy.get("#__cy_root > :nth-child(1) > :nth-child(3)").should(
+    cy.get("[data-cy=submissionDateTop]").should("exist");
+    cy.get("[data-cy=submissionDateTop]").should(
       "include.text",
       "Form Submitted on: 02/07/2022"
     );
-    cy.get("#__cy_root > :nth-child(1) > :nth-child(7)").should("exist");
-    cy.get("#__cy_root > :nth-child(1) > :nth-child(7)").should(
+    cy.get("[data-cy=submissionDate]").should("exist");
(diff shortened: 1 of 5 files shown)


src/components/forms/formr-part-b/viewSections/View.tsx
```diff
@@ -57,7 +58,6 @@ const View = ({ canEdit, history }: IView) => {
 
   const subDateContent = `Form Submitted on:
```
john12321, line 59: This is not needed now.

src/utilities/FormRUtilities.tsx
```diff
@@ -117,6 +119,17 @@ export class FormRUtilities {
     return value ? value : message;
   }
 
+  public static dispaySubmissionDate(date: DateType) {
```
john12321, line 122: You've got an unused `date` parameter here.

src/utilities/FormRUtilities.tsx
```diff
@@ -117,6 +119,17 @@ export class FormRUtilities {
...
+    const message = (
+      <h3>
+        {"Form Submitted on: " +
+          DateUtilities.ToLocalDate(formData.submissionDate)}
```
john12321, line 126: ... the `date` parameter goes inside the `ToLocalDate` method rather than the argument (the actual value passed when we use the util method in the components).

src/utilities/FormRUtilities.tsx
john12321, line 122: Also, I think you have a typo in util method name:  `dispaySubmissionDate` (missing an 'l'?)

john12321 (changes requested):

src/components/forms/formr-part-a/View.cy-test.tsx
```diff
@@ -53,10 +53,16 @@ describe("View", () => {
     cy.get("[data-cy=BtnEdit]").should("not.exist");
     cy.get("[data-cy=BtnSaveDraft]").should("not.exist");
     cy.get("[data-cy=BtnSubmit]").should("not.exist");
-    cy.get("[data-cy=submissionDateTop]").should("exist");
```
john12321, line 56: Sorry to be picky, Ed but I think it's better to keep the `data-cy=submissionDateTop` tags etc. you had earlier rather than `#__cy_root > :nth-child(1) > :nth-child...` to keep it isolated from any changes we might make - and it's also easier to read!

john12321 (approved):

PR #733, opened 2022-10-21
##### DRAFT ###### run a matrix of tests using electron, chrome and firefox

johnsimmons1980:
Not sure if this is really needed, (or if it'll work ...) but it may be good to see how it goes running the e2e tests on different browsers, this PR will set cypress to use electron, chrome and firefox, but if chrome or firefox crash and electron complete successfully then it will just continue to the next step of the deploy pipeline, only if electron fails its tests then we fail that job.

🤞🏻

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,34 +7,45 @@ on:
   workflow_dispatch:
 
 jobs:
+  continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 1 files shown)


john12321 (approved):
I like experiments!
It would be good if we could run the test files in parallel too to save time lol!

DorisWongNHS (approved):

PR #740, opened 2022-10-27
GHA to check for dependencies before applying

johnsimmons1980:
this action should allow us to tag PR's with the "Dependencies" label, and add the following text to the PR:


Judge40:
The used action is no longer supported

PR #763, opened 2022-11-24
Refactor Profile section

john12321:
TIS21-SHED

A bit of personal development/ pre-DSP beta work to make Profile a bit DRYer via a PanelsCreator comp.
Not fully generic due to nested Curricula array and Personal Details not included in the generator yet to keep the Profile UI/UX as-is.
New Profile Panels use the latest NHS react comps while again trying to keep the Profile panel look as-is.
More Cypress CT added for those cases where there's no panel data to display.
Keeping the generator component as generic as possible requires a couple more functions to order, filter and format the panel data depending on what type of panel is required. Next steps could include making things a bit more generic so they can be more easily reused throughout TSS. 

ps. Added a genericSort method as a better fix for some code smells.


src/redux/slices/traineeProfileSlice.ts
```diff
@@ -52,7 +53,17 @@ const traineeProfileSlice = createSlice({
       })
       .addCase(fetchTraineeProfileData.fulfilled, (state, action) => {
         state.status = "succeeded";
-        state.traineeProfileData = action.payload;
+        state.traineeProfileData.traineeTisId = action.payload.traineeTisId;
+        state.traineeProfileData.personalDetails =
+          action.payload.personalDetails;
+        state.traineeProfileData.programmeMemberships =
+          action.payload.programmeMemberships;
```
Judge40, line 60: Do we not need to sort programme memberships too?

Judge40 (approved):

src/redux/slices/traineeProfileSlice.ts
john12321: @Judge40 Good shout. Did leave as-is because couldn't find any sort on the Programmes data in the current code.  Either way, would be good to get some more use out of the genericSort. (change could even be as a v small follow-up PR! 😁)

Judge40 (approved):

PR #768, opened 2022-12-06
Update React and CRA

john12321:
Update: 
- React v17 to 18
- react-scripts (CRA) v4 to 5
- Cypress v8 to 11
- Redo Autocomplete comp (MUI breaking change)
- Remove Enzyme tests (breaking change)

Issues:
1. Can't instrument the Cypress CT's using @cypress/instrument-cra  - does not appear to be well maintained for react-scripts 5. Ideally should be using Istanbul plugin directly (which CRA doesn't allow without hacking/ejecting from CRA) - which we can do when migrated to Vite/ NEXT JS.
2. There are a few errors (that don’t appear to break anything - yet) relating to Polyfilling node core modules. CRA 4 used to include NodeJS polyfills by default but not anymore. Again, best course of action is migration to Vite/Next JS.

TODO:
Create follow-up tickets for:
1. Use Cypress to redo some of the deleted Jest Enzyme tests (where not covered by Cypress).
2. Update as many other libraries as possible before migration to Vite/Next JS.

Then escape from CRA!

### DIFF
--- a/cypress/component/forms/SubmittedFormsList.cy.tsx
+++ b/cypress/component/forms/SubmittedFormsList.cy.tsx
@@ -9,7 +9,6 @@ import {
   recentSubmitDate
 } from "../../../src/mock-data/submitted-formr-parta";
 import history from "../../../src/components/navigation/history";
-import React from "react";
 
 describe("SubmittedFormsList", () => {
   it("should show no forms submitted msg if no submitted forms", () => {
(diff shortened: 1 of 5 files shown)


src/browser-update-config.json
```diff
@@ -1,23 +1,21 @@
 {
   "required": {
-    "e": -2,
-    "f": -3,
-    "c": -3,
-    "o": -3,
```
Judge40, line 6: What's the default behaviour after removing these?

john12321: Docs talk about falling back to a default (was assuming they would have a sensible one?). Happy to be more specific though (main worry was that damn Safari!).

Judge40 (approved):

src/browser-update-config.json
john12321: Will have a go on BrowserStack to see how sensible they are!

PR #775, opened 2022-12-12
fix: formr part a comp

john12321:
Remove the programmeSpecialty conditional mistakenly added. 
Refactor AutocompleteSelect comp so it can handle undefined value (if needed).

NO TICKET


ReubenRobertsHEE (approved):
Here y'are :)

PR #787, opened 2023-01-18
Feat/dockerfile

ReubenRobertsHEE:


john12321 (approved):

ReubenRobertsHEE:
The environment variables are hardcoded here - these will need to be set to pick up from the proper environment variables before we push to prod, but I'm assuming this is intended to go no further than staging at this point!

PR #790, opened 2023-01-26
fix: rename static html file

john12321:
Cloudwatch can't handle [[...app]].html format.

NO-TICKET

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -33,6 +33,7 @@ jobs:
         run: npm install && npm run build
 
       - name: Deploy app build to S3 bucket
+        run: mv ./out/[[...app]].html ./out/index.html
         run: aws s3 sync ./out/ s3://${{ secrets.AWS_S3_BUCKET_PREPROD }} --delete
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         run: npm install && npm run build
 
       - name: Deploy app build to S3 bucket
+        run: mv ./out/[[...app]].html ./out/index.html
```
Judge40, line 36: You'll need to format this as a multi-line command
e.g.
```
run: |
  mv ./out/[[...app]].html ./out/index.html
  aws s3 sync ./out/ s3://${{ secrets.AWS_S3_BUCKET_PREPROD }} --delete
```

Judge40 (changes requested):

.github/workflows/deploy.yml
john12321: ah nice one @Judge40 🌮

Judge40 (approved):

PR #793, opened 2023-01-27
fix: replace environment API call with env files

Judge40:
The environment API call was implemented to support a "build once, run anyway" approach using Docker containerisation. The approach is non-standard and causes issues with the changes to migrate to Next.js.

Replace the environment API call with dotenv files. To continue supporting both preprod and prod a custom build script should be used to swap the correct dotenv contents in depending on the environment.

TIS21-3995


john12321:
LGTM!

john12321 (approved):

PR #806, opened 2023-02-15
chore: set NHS font-family

john12321:
NO TICKET

### DIFF
--- a/styles/globals.scss
+++ b/styles/globals.scss
@@ -6,17 +6,13 @@ $nhsFont: “Frutiger W01”, Arial, sans-serif;
 
 html {
   background-color: inherit;
+  font-family: "Frutiger W01", Arial, sans-serif !important;
 }
 
 body {
   margin: 0;
   -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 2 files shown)


Judge40 (approved):

styles/globals.scss
```diff
@@ -6,6 +6,7 @@ $nhsFont: “Frutiger W01”, Arial, sans-serif;
 
 html {
   background-color: inherit;
+  font-family: "Frutiger W01", Arial, sans-serif !important;
```
ReubenRobertsHEE, line 9: `!important ` really? 😁

john12321: @ReubenRobertsHEE don't take away my comfort blanket! 😅
Default Amplify css has had a habit of overriding things so was taking no chances (another PR to fix 😁).
Happy to remove if you're feeling lucky?

DorisWongNHS (approved):

styles/globals.scss
ReubenRobertsHEE: Definitely not, I'm all for comfort blankets and such like 😀 ⛑️

PR #850, opened 2023-04-25
fix: signCoj service function passing null body

Judge40:
The `TraineeProfileService.signCoj()` function passes `null` for the POST request body, but that is not allowed.
Update the `ApiService` to allow the POST request form data to be optional.

TIS21-4389
TIS21-4374


DorisWongNHS (approved):

DorisWongNHS:
Sorry I and got this updated on my branch, I should have updated it on Ed's branch earlier.

PR #851, opened 2023-04-25
Feat/add coj notification banner

Judge40:
Show an alert at the top of each page if the user has any Programme Memberships with a Conditions of Joining available to sign.

Some of the component/styling varies from the POC, some reasons are explained in inline PR comments.
Component tests to follow for both the banner behaviour and COJ status field.

TIS21-4373

### DIFF
--- a/components/main/GlobalAlert.tsx
+++ b/components/main/GlobalAlert.tsx
@@ -1,0 +1,42 @@
+import { NavLink, useLocation } from "react-router-dom";
+import { useAppSelector } from "../../redux/hooks/hooks";
+
+const GlobalAlert = () => {
+  return (
+    <div
+      className="app-global-alert"
+      id="app-global-alert"
+      role="complementary"
+    >
+      <div className="nhsuk-width-container">
+        <CojAlert />
+      </div>
+    </div>
+  );
+};
+
+function CojAlert() {
+  const pathName = useLocation().pathname;
+  const showCojAlert = useAppSelector(
+    state => state.traineeProfile.hasSignableCoj
+  );
+
+  return !showCojAlert ? null : (
+    <div className="nhsuk-grid-row">
+      <div className="nhsuk-grid-column-full">
+        <div className="app-global-alert__content">
+          <div className="app-global-alert__message">
+            <h2>Please sign your Conditions of Joining Agreement</h2>
+            {pathName === "/programmes" ? null : (
+              <NavLink data-cy="cojLink" to={`/programmes`}>
+                <p>Click here to navigate to your Programmes page to sign</p>
+              </NavLink>
+            )}
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export default GlobalAlert;
--- a/components/main/Main.tsx
+++ b/components/main/Main.tsx
@@ -1,4 +1,4 @@
-import { Switch, Route, Redirect, Router } from "react-router-dom";
+import { Switch, Route, Redirect, Router, useLocation } from "react-router-dom";
 import Profile from "../profile/Profile";
 import FormRPartA from "../forms/formr-part-a/FormRPartA";
 import FormRPartB from "../forms/formr-part-b/FormRPartB";
@@ -24,6 +24,7 @@ import Placements from "../placements/Placements";
 import Programmes from "../programmes/Programmes";
 import Breadcrumbs from "../breadcrumbs/Breadcrumbs";
 import Dsp from "../dsp/Dsp";
+import GlobalAlert from "./GlobalAlert";
 
 interface IMain {
   signOut: any;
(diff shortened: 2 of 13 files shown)


utilities/Constants.ts
```diff
@@ -22,7 +22,9 @@ export const MFA_OPTIONS = [
   }
 ];
 
-export const COJ_EPOCH = new Date("2023-08-01");
+export const COJ_EPOCH = new Date(
+  process.env.NEXT_PUBLIC_CONDITIONS_OF_JOINING_EPOCH as string
+);
```
Judge40, line 27: It feels a bit awkward having a "constant" set via env vars, but we have a real issue when it comes to testing both locally and on stage. This allows setting the "COJ epoch" per environment and allows us to verify the stage deployment without any risk of enabling COJ on prod.

styles/globals.scss
```diff
@@ -57,6 +57,14 @@ body {
   text-align: center;
 }
 
+.app-global-alert {
+  background-color: #ffeb3b;
+}
+
+.app-global-alert__content {
+  padding-top: 24px;
```
Judge40, line 65: This will need tweaking if we add another type of alert, otherwise the padding between alerts is too much. It was added here instead of `.app-global-alert` to avoid a thin yellow strip being displayed when there was no `<CojAlert>` element.

components/main/Main.tsx
```diff
@@ -73,51 +74,62 @@ export const Main = ({ signOut, appVersion }: IMain) => {
...
-                to={{
-                  pathname: "/credential/invalid",
-                  search: location.search
-                }}
-              />
-              <Route path="/*" component={PageNotFound} />
-            </Switch>
-          </main>
-          <HEEFooter appVersion={appVersion} />
+          <App signOut={signOut} appVersion={appVersion} />
```
Judge40, line 77: This needed to be moved to a function to allow `useLocation()` to work within the `<Router>`

components/main/GlobalAlert.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const GlobalAlert = () => {
+  return (
+    <div
+      className="app-global-alert"
+      id="app-global-alert"
+      role="complementary"
+    >
+      <div className="nhsuk-width-container">
+        <CojAlert />
```
Judge40, line 12: The `GlobalAlert` is always rendered, the `CojAlert` may or may not be rendered based on the current state. If no alerts are rendered then GlobalAlert will still exist in the DOM but not be displayed.

The decision to render `CojAlert` was deferred to itself rather than logic here so that we could avoid complicating the GlobalAlert logic if/when further types of alerts are added.

Judge40:
The current easiest way to test the banner is to set up your data to have a signable COJ, then use the Redux dev tools to dispatch an event like below to "sign" the COJ and update the banner state.

```
{
  type: 'traineeProfile/programmeMembership/signCoj/fulfilled',
  payload: {
      tisId: '12345',
      programmeTisId: '1',
      programmeName: 'Cardiology',
      programmeNumber: 'EOE8945',
      managingDeanery: 'East of England',
      programmeMembershipType: 'SUBSTANTIVE',
      startDate: '2023-05-01',
      endDate: '2028-01-01',
      programmeCompletionDate: '2023-12-31',
      status: 'FUTURE',
      curricula: [
        {
          curriculumTisId: '14',
          curriculumName: 'GP Returner',
          curriculumSubType: 'MEDICAL_CURRICULUM',
          curriculumStartDate: '2019-01-01',
          curriculumEndDate: null
        },
        {
          curriculumTisId: '2',
          curriculumName: 'ST2',
          curriculumSubType: 'ACF_OTHER_FUNDING',
          curriculumStartDate: '2019-06-01',
          curriculumEndDate: null
        }
      ],
      conditionsOfJoining: {
        signedAt: '2023-02-02T00:00:00Z',
        version: 'GG9'
      },
      signature: {
        signedAt: '2023-04-21T15:24:35.398166500Z',
        validUntil: '2023-04-22T15:24:35.398166500Z',
        hmac: '20bf844a4105d0121c3f5fc382dc618431e3049f4860896e19092bfe6538cc00'
      }
    }
}
```

components/main/GlobalAlert.tsx
john12321: Yeah , I see your thinking here to make the GlobalAlert more generic. Maybe we could conditionally render it by e.g. checking an array of functions e.g. showCojAlert, showNextThingAlert etc.., if any are true then GlobalAlert is rendered along with any child alerts that were true?

john12321 (approved):

Judge40:
Merging as is, with an agreement to refactor the GlobalAlert rendering in a follow up PR

PR #853, opened 2023-04-26
refactor: GlobalAlert rendering

Judge40:
Refactor the rendering condition of GlobalAlert so that it is not rendered as a zero height element if the CojAlert is not rendered. This makes it more extensible in the future, where multiple alert types may be available.

TIS21-4454
TIS21-4373


john12321 (approved):
Haven't pulled your branch but LGTM from here!

PR #854, opened 2023-04-27
Feat/sign coj

DorisWongNHS:


components/forms/conditionOfJoining/CojView.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <>
+      <Formik
+        initialValues={{}}
+        onSubmit={async _values => {
+          const traineeProfileService = new TraineeProfileService();
+          const response: AxiosResponse<ProgrammeMembership> =
+            await traineeProfileService.signCoj(
+              store.getState().user.signingCojPmId
+            );
+          store.dispatch(updatedSignedCoj(response.data.conditionsOfJoining));
```
Judge40, line 45: There is already a thunk to handle the service call and updating of the state https://github.com/Health-Education-England/tis-trainee-ui/blob/main/redux/slices/traineeProfileSlice.ts#L39

components/main/Main.tsx
```diff
@@ -82,6 +83,7 @@ export const Main = ({ signOut, appVersion }: IMain) => {
             <Route exact path="/home" component={Home} />
             <Route exact path="/placements" component={Placements} />
             <Route exact path="/programmes" component={Programmes} />
+            <Route exact path="/programmes/:id" component={CojView} />
```
Judge40, line 86: The route should indicate that it is for signing coj e.g. `/programme/:id/sign-coj`

Judge40 (changes requested):
I'm commenting as I unpack the implementation, so this isn't a complete review yet.

Judge40 (approved):

components/forms/conditionOfJoining/CojView.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useEffect } from "react";
+import { signCoj } from "../../../redux/slices/traineeProfileSlice";
+import { updatedsigningCojProgName } from "../../../redux/slices/userSlice";
+import store from "../../../redux/store/store";
+import history from "../../navigation/history";
+import ScrollTo from "../ScrollTo";
+import CojGg9 from "./CojGg9";
+
+const CojView: React.FC = () => {
+  const progName = store.getState().user.signingCojProgName;
```
john12321, line 12: This is more of TODO item to look at a for future ticket than a request for changes. 
We need to look to see if we we are making best use of the `signingCojProgName` state variable to track the user's progress in signing the form. The original  idea was to prevent users from bypassing the CojView btn onSubmit. And the `return` statement in the `useEffect` resetting the `signingCojProgName` to null when leaving the page.

john12321 (approved):

john12321:
Nice work, Doris! 🌮

PR #859, opened 2023-05-02
Feat/coj agreements

DorisWongNHS:


components/forms/conditionOfJoining/CojView.tsx
```diff
@@ -27,22 +31,125 @@ function CojDeclarationSection() {
...
+              id="isDeclareEngage"
+              type="checkbox"
+              name="isDeclareEngage"
+              items={[
+                {
+                  label: `Where applicable, I will fully engage with immigration and employer requirements relating to Tier 2 and Tier 4 UK visas.`,
+                  value: true
+                }
+              ]}
+            />
```
john12321, line 146: @DorisWongNHS , the implementation looks good. One thing you might want to do is refactor this component (following DRY principle). e.g. make an array of checkboxProps objects (key/value pairs of id, label props) which can be imported from the constants file), then loop through this array to render a MultiChoiceInputField for each item in the array.

john12321 (changes requested):
Looks good, @Doris 🌮  Just left one comment to look at some minor refactoring of the CojView component.

components/forms/conditionOfJoining/CojView.tsx
```diff
@@ -66,84 +66,19 @@ function CojDeclarationSection() {
...
-              name="isDeclareEngage"
-              items={[
-                {
-                  label: `Where applicable, I will fully engage with immigration and employer requirements relating to Tier 2 and Tier 4 UK visas.`,
-                  value: true
-                }
-              ]}
-            />
+            {COJ_DECLARATIONS.map(declaration => (
+              <MultiChoiceInputField
```
john12321, line 70: Looks great, @DorisWongNHS . Just one small error (check your console when you render the CojView.tsx. You should see a warning something like this: 
`each child should have a unique "key" prop.`). 
This can be fixed by wrapping the MultiChoiceInputField in a React.Fragment and giving it key prop with a unique value e.g. declaration.id.
Sorry for being picky! 😅

john12321 (changes requested):

john12321 (approved):
Nice work, @DorisWongNHS! 🌮

PR #861, opened 2023-05-03
fix: alerts object

john12321:
Wrap the alerts object in a useMemo hook to prevent unnecessary re-renders. At the moment, it makes the dependencies of the useEffect change on every render - which is not good for performance. It will now only be re-evaluated when the showCojAlert value changes.

JFDI

### DIFF
--- a/components/main/GlobalAlert.tsx
+++ b/components/main/GlobalAlert.tsx
@@ -1,5 +1,5 @@
 import { NavLink, useLocation } from "react-router-dom";
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import React from "react";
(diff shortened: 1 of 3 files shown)


ReubenRobertsHEE:
To be reviewed by trainee backend, eh? #fullstack-for-the-win 😁

ReubenRobertsHEE:
But... but... where are the TESTS?! :trollface:

john12321:
> But... but... where are the TESTS?! :trollface:

🤣 @ReubenRobertsHEE Time for some of those perf tests again! (failing that: has the eslint warning gone?! #ineslintwetrust)

ReubenRobertsHEE (approved):

PR #862, opened 2023-05-03
fix: Coj contents details

DorisWongNHS:


styles/globals.scss
```diff
@@ -374,3 +374,7 @@ button:disabled {
     display: table-cell !important;
   }
 }
+
+ol[type="a"] {
```
john12321, line 378: I like this @DorisWongNHS !

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.65.1",
+  "version": "0.65.3",
```
john12321, line 3: this might be a typo but no big deal.

john12321 (approved):

package.json
DorisWongNHS: I am not sure the order we want to merge it. But @Judge40's PR is having `0.65.2`
https://github.com/Health-Education-England/tis-trainee-ui/pull/860/files

john12321: Ah OK, I think you may have got there first! 😄

ReubenRobertsHEE (dismissed):

john12321 (approved):

PR #863, opened 2023-05-05
feat: showCoj stuff to prod users

john12321:


ReubenRobertsHEE:
Naz asked to do a bit more testing before we merge this. After the school run 🤞

PR #866, opened 2023-05-11
fix: COJ agreements wording

DorisWongNHS:


Judge40 (commented):
I found more differences than these 🤔

components/forms/conditionOfJoining/CojGg9.tsx
```diff
@@ -173,7 +178,7 @@ const CojGg9: React.FC<CojGg9Props> = ({ progName }) => {
               <li>
                 I will maintain regular contact with my Training Programme
                 Director, other trainers and HEE/NES/HEIW/NIMDTA by responding
-                promptly to communications from them.
+                promptly to communications from them
```
Judge40, line 181: This does have a fullstop in the GG version

utilities/Constants.ts
```diff
@@ -30,7 +30,7 @@ export const COJ_DECLARATIONS = [
   {
     id: "isDeclareProvisional",
     label:
-      "I understand that programme and post allocations are provisional and subject to change until confirmed by HEE WM and/or my employing organisation."
+      "I understand that programme and post allocations are provisional and subject to change until confirmed by HEE/NES/ HEIW/NIMDTA and/or my employing organisation."
```
Judge40, line 33: The GG does have a space between `NES` and `NEIW` but that seems like an error compared to other usages, so we can remove that extra whitespace here.

Judge40 (changes requested):
There are a couple of places I think the GG is actually formatted wrong, but otherwise this all looks good.

Judge40 (approved):

PR #870, opened 2023-05-19
add text link for viewing ReadOnly form

EdwardBarclay:
TIS21-4534

### DIFF
--- a/components/programmes/ConditionsOfJoining.tsx
+++ b/components/programmes/ConditionsOfJoining.tsx
@@ -41,18 +41,15 @@ export function ConditionsOfJoining({
         </SummaryList.Value>
       </SummaryList.Row>
       <SummaryList.Row>
-        <SummaryList.Value
-          onClick={() => viewCoj(programmeMembershipId, programmeName)}
-          className="plain-text-link"
-          data-cy={`cojViewBtn-${programmeMembershipId}`}
-          style={{
-            textDecoration: "underline",
-            cursor: "pointer",
-            whiteSpace: "nowrap"
-          }}
-        >
-          View Signed Condition of Joining
-        </SummaryList.Value>
+        <SummaryList.Actions style={{ borderBottom: 0 }}>
+          <Button
+            secondary
+            onClick={() => viewCoj(programmeMembershipId, programmeName)}
+            data-cy={`cojViewBtn-${programmeMembershipId}`}
+          >
+            View
+          </Button>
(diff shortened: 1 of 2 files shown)


components/programmes/ConditionsOfJoining.tsx
```diff
@@ -40,6 +40,20 @@ export function ConditionsOfJoining({
...
+          onClick={() => viewCoj(programmeMembershipId, programmeName)}
+          className="plain-text-link"
+          data-cy={`cojViewBtn-${programmeMembershipId}`}
+          style={{
+            textDecoration: "underline",
+            cursor: "pointer",
+            whiteSpace: "nowrap"
+          }}
+        >
+          View Signed Condition of Joining
```
Judge40, line 54: We're already in the context of a signed conditions of joining, so this can just be a simple "View"

components/programmes/ConditionsOfJoining.tsx
```diff
@@ -40,6 +40,20 @@ export function ConditionsOfJoining({
           {CojUtilities.getVersionText(conditionsOfJoining.version)}
         </SummaryList.Value>
       </SummaryList.Row>
+      <SummaryList.Row>
+        <SummaryList.Value
```
Judge40, line 44: This really should be a button within a `SummaryList.Actions` to stay consistent with the "Sign" button.

Judge40 (changes requested):
This is stepping on the toes of TIS21-4544 and creates inconsistency with the Sign button.
You could drop the `SummaryList.Actions` and child `Button` from the unsigned coj part of the component and just switch `cojSignBtn` to `cojViewBtn` and the text from "Sign" to "View" as a quick implementation with consistency.

cypress/component/programmes/Programmes.cy.tsx
```diff
@@ -282,4 +282,36 @@ describe("Programme summary panel", () => {
...
+          <MockedProgrammes />
+        </Router>
+      </Provider>
+    );
+
+    cy.get("[data-cy='cojSignedDate']").should("exist");
+
+    cy.get("[data-cy='cojViewBtn-1']")
+      .should("exist")
+      .and("have.text", "View Signed Condition of Joining");
```
Judge40, line 315: This test will fail given the changes to the button text

Judge40 (changes requested):

Judge40 (approved):

PR #873, opened 2023-05-22
refactor: style COJ actions as links

Judge40:
The COJ actions are currently a button, but should be styled as links instead.
Swap the `button` element for `Link`s and refactor the `viewCoj` function in to `setCojState` with the navigation done within the link itself.

Update the ConditionOfJoining component tests to wrap the component in a `BrowserRouter` so that the `Link`s can work during testing.

TIS21-4544
TIS21-4533

### DIFF
--- a/components/programmes/ConditionsOfJoining.tsx
+++ b/components/programmes/ConditionsOfJoining.tsx
@@ -1,4 +1,5 @@
-import { Button, SummaryList } from "nhsuk-react-components";
+import { SummaryList } from "nhsuk-react-components";
+import { Link } from "react-router-dom";
 import { ConditionsOfJoining as ConditionsOfJoiningModel } from "../../models/ProgrammeMembership";
 import {
   updatedsigningCoj,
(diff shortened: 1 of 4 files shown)


components/programmes/ConditionsOfJoining.tsx
```diff
@@ -42,13 +42,13 @@ export function ConditionsOfJoining({
       </SummaryList.Row>
       <SummaryList.Row>
         <SummaryList.Actions style={{ borderBottom: 0 }}>
-          <Button
-            secondary
-            onClick={() => viewCoj(programmeMembershipId, programmeName)}
+          <Link
+            to={`/programmes/${programmeMembershipId}/sign-coj`}
+            onClick={() => setCojState(programmeMembershipId, programmeName)}
```
Judge40, line 47: Splitting the state update and navigation appears to work fine, am I correct in thinking the `onClick` will always complete before the navigation?

EdwardBarclay (approved):

PR #877, opened 2023-05-31
Create a form builder and use it to create Form R Part A 

john12321:
Create a FormBuilder component that builds Form R Part A via a JSON file.

The FormBuilder includes the following:  
- All fields with errors will be listed in the error summary box (currently blank).
- Fields that are hidden and then unhidden do not have their value reset (but the hidden data is not validated and not included in submitted form data).
- Safari-compatible react-datepicker is used for FormBuilder date type (native date type currently breaks app if Safari <15).
- react-phone-number-input comp (already used in MFA SMS set-up) is used for FormBuilder phone type.

TODO
- Complete the use of localStorage for saving and retrieval of Form R Part A (needs BE changes to trainee-forms DTO which will be done in a follow-up ticket).
- Create Work panel and Total FormBuilder types (used in Form R (Part B).
- Build Form R (Part B) via FormBuilder comp/ JSON file.


EdwardBarclay (approved):

Judge40 (approved):
I didn't do a full line-by-line review, but the overall architecture is fairly easy to follow.

The QoL improvements are really nice in terms of the new input fields and the validation not instantly spamming red failures the moment you touch the form.

The [dev-handbook](https://github.com/Health-Education-England/dev-handbook/blob/ad16b834143886ef09b1e9de3d19df4ba0975dc4/tis-self-service/environment-setup/initialization/db/create-reference-dummy-data.js#L184) will need updating to make sure that the `curriculumSubType` is set, I couldn't complete the form at first because I couldn't select a Programme Specialty. Other than that, really nice work @john12321 🌮

john12321:
Cheers for the late Friday review & feedback @Judge40 💪! 
I've just spotted a bug where the form fields built via the JSON file are not resetting to their original state when a new form is opened (e.g. changes made to unhide a field are persisting when navigating away from the form then opening a new one). I've got a fix for this which I will push in a minute.

Judge40 (approved):

PR #880, opened 2023-06-06
fix: 2 formBuilder bugs

john12321:
Fixed (hopefully without breaking anything else) bugs:
1. JSON form fields' visibility were not reset when navigating away then opening a new form. This was fixed by not directly mutuating the (JSON forms) fields.

2. When completing a form and making a dependent field visible via a click event on the parent (e.g. programme specialty radio click unhides the 2 cct fields), these fields should remain visible and therefore subject to validation. However, a bug meant that these dependent fields' visibility was reset to the default hidden when the saved draft form was reopened so the validation did not fire when navigating to the next page. This was fixed by Initialising the dependent (JSON form) fields' visibility based on their parent field's (fetchedFormData) value/state.

TODO: Added  an e2e test for the first fix but need to work out a way to set up Cypress to test the first. (follow-up ticket)

ps. Haven't got the will to fix that code smell today!

### DIFF
--- a/components/forms/FormsListBtn.tsx
+++ b/components/forms/FormsListBtn.tsx
@@ -17,7 +17,7 @@ import {
   resetLocalStorageFormData
 } from "../../utilities/FormBuilderUtilities";
 import { LifeCycleState } from "../../models/LifeCycleState";
-import { useEffect } from "react";
+import { useCallback, useEffect } from "react";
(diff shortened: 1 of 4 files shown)


ReubenRobertsHEE (approved):
I'm happy to give this a decent going over on staging. I haven't looked at all the details of the fix.

PR #905, opened 2023-07-19
Add 'What's New' to TSS Home Page

john12321:
- Add a 'What's New' section on Home next to Welcome
- Display summary of 1 most recent post in this section
- Show 'no posts' message and 'couldn't show any posts' messages in this section for empty and error states respectively.

TIS21-4528

### DIFF
--- a/components/home/TssUpdates.tsx
+++ b/components/home/TssUpdates.tsx
@@ -7,36 +7,36 @@ type PostSummary = Pick<Post, "id" | "title" | "excerpt">;
 
 export const TssUpdates: React.FC = () => {
   const whatsNewHeader = "What's New";
+  const wpUpdatesUrl =
+    "https://tis-support.hee.nhs.uk/about-tis/welcome-to-the-tss-updates/";
   const whatsNewStatus = useAppSelector(state => state.tssUpdates.status);
   const whatsNewPosts: Post[] = useAppSelector(
     state => state.tssUpdates.tssUpdates
   );
-
-  const filteredPosts = useMemo(() => {
-    const filteredPosts: PostSummary[] = [];
-    if (whatsNewPosts.length === 0) {
-      return [];
-    }
-    for (const post of whatsNewPosts) {
-      if (post.tags.includes(20)) {
-        filteredPosts.push({
+  const newSummaryPosts = useMemo(() => {
+    const summaryPosts: PostSummary[] = [];
+    if (whatsNewPosts?.length > 0) {
+      for (const post of whatsNewPosts) {
+        summaryPosts.push({
           id: post.id,
           title: post.title,
           excerpt: post.excerpt
         });
       }
-    }
-    return filteredPosts;
+    } else return [];
+    return summaryPosts;
   }, [whatsNewPosts]);
 
   const addWhatsNewPosts = (): JSX.Element[] | JSX.Element => {
-    if (filteredPosts.length === 0) {
-      return <p data-cy="noUpdates">No updates available</p>;
+    if (newSummaryPosts.length === 0) {
+      return <p data-cy="noUpdates">No new updates available at the moment.</p>;
     }
-    return filteredPosts.map(post => {
+    return newSummaryPosts.map(post => {
       return (
         <ListPanel.Item key={post.id}>
-          <h3 data-cy={`postTitle${post.id}`}>{post.title.rendered}</h3>
+          <h3 className="list-panel-header" data-cy={`postTitle${post.id}`}>
+            {post.title.rendered}
+          </h3>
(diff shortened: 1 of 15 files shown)


components/home/TssUpdates.tsx
```diff
@@ -0,0 +1,102 @@
...
+          excerpt: post.excerpt
+        });
+      }
+    }
+    return filteredPosts;
+  }, [whatsNewPosts]);
+
+  const addWhatsNewPosts = (): JSX.Element[] | JSX.Element => {
+    if (filteredPosts.length === 0) {
+      return <p data-cy="noUpdates">No updates available</p>;
```
ReubenRobertsHEE, line 34: Tested - works correctly

components/home/TssUpdates.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+    return filteredPosts.map(post => {
+      return (
+        <ListPanel.Item key={post.id}>
+          <h3 data-cy={`postTitle${post.id}`}>{post.title.rendered}</h3>
+          <p data-cy={`postExcerpt${post.id}`}>
+            {extractTextFromHTML(post.excerpt.rendered)}
```
ReubenRobertsHEE, line 41: Tested on post with no manual excerpt - wordpress generates an automatic one itself, so this is fine.

ReubenRobertsHEE:
I've noticed the styling goes a bit funny with some short-height dimensions: not sure if this is intentional?

whats_new_tall
whats_new_short

services/TssUpdatesService.ts
```diff
@@ -0,0 +1,5 @@
+const wpUrl = "https://tis-support.hee.nhs.uk/wp-json/wp/v2/posts?per_page=20";
```
ReubenRobertsHEE, line 1: (From chat):
Here's an API call which works I think:
https://tis-support.hee.nhs.uk/wp-json/wp/v2/posts?categories=19&orderby=date&status=publish&per_page=3
The wordpress page has to use the category to select what posts to display though :sadcowboy:  - so it might avoid confusion in future if we use the same on the API call

components/home/TssUpdates.tsx
```diff
@@ -0,0 +1,102 @@
...
+  );
+
+  const filteredPosts = useMemo(() => {
+    const filteredPosts: PostSummary[] = [];
+    if (whatsNewPosts.length === 0) {
+      return [];
+    }
+    for (const post of whatsNewPosts) {
+      if (post.tags.includes(20)) {
+        filteredPosts.push({
```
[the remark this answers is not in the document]
john12321: Using your suggested category filter API will stop this nonsense!🌮

john12321:
> I've noticed the styling goes a bit funny with some short-height dimensions: not sure if this is intentional?
> 
> whats_new_tall whats_new_short

Good spot. will tweak the CSS to make it more user-friendly for those landscape mobile users. Cheers!

ReubenRobertsHEE:
> @ReubenRobertsHEE Great feedback/ API updates. Now we're not up against the (previous) Review deadline, still having thoughts about the nested scrollbar. Thinking to keep 3 posts for now and have a play on browserstack to see how things behave unless you have a strong opinion the other way? Thanks again and see you at the next meeting!

Sounds good to me. You know, if we wanted to go Super Saiyan on this, we could look at those dismissable speech-bubble type popups that you get on some sites to highlight something new, or to walk you through a process for the first time 😮‍💨

john12321:
> > @ReubenRobertsHEE Great feedback/ API updates. Now we're not up against the (previous) Review deadline, still having thoughts about the nested scrollbar. Thinking to keep 3 posts for now and have a play on browserstack to see how things behave unless you have a strong opinion the other way? Thanks again and see you at the next meeting!
> 
> Sounds good to me. You know, if we wanted to go Super Saiyan on this, we could look at those dismissable speech-bubble type popups that you get on some sites to highlight something new, or to walk you through a process for the first time 😮‍💨

😁 Dragon Ball Z it! Nice. Yeah I really like that idea. Also ties in well with that granular feedback at the element level I was going on about yesterday. These 'bubbles' would be good to have next to a change together with a quick '💩to 😎' rating which we could capture on a nice little Node server 🤣 .

ReubenRobertsHEE (approved):

ReubenRobertsHEE:
If we just show a single 'what's new', we can queue-up the posts we've already written by converting all but the oldest to draft, and then publishing them one by one every week or two. That way they will each get their dose of limelight ;)

john12321:
I’ll set my ‘replace the stale post’ alarm! 😁 Haven’t touched WP for ages, don’t suppose there’s a script (PHP !) to archive, delay publish etc. ?

PR #906, opened 2023-07-21
Chore/remove redundant what's new code

john12321:
Given that we're sticking to the AC's 😁 and only showing the latest 'what's new' post, we can ditch some useless TS / CSS.
Also tidied up the styling for when there is no post to show.

NO TICKET


ReubenRobertsHEE (approved):
Cool, nice Johno. We can always fish some of this code out of the git history if we end up flip-flopping and showing multiple news items!

PR #930, opened 2023-08-18
Feat/add autosave and 'start over' btn

john12321:
Main changes I can think of:

- Add autosave to form r 
1. Autosave triggered 2s after the user pauses making changes (we can always tweak this if users find the update flicker annoying!)
2. Displayed autosave message/timestamp depends on autosaveStatus 
3. No draft form if no changes (tested via making no form changes and then navigating away from form and then back again)
4. Autosave will continue if change made followed by quick nav to another form section

Trickiest bit was keeping a source of truth for the form following first autosave (from NEW to DRAFT with id etc.). The response from the autosave POST/PUT came in handy!

- Add 'start over' button (allows draft form deletion)
1. 'Start over' button is visible on form page after first autosave success
2. Button will be visible on CreateList (forms list) page if DRAFT/UNSUBMITTED form in list.
3. Successful draft form deletion via btn click on form page will display success toast and redirect to CreateList where 'submit new form' btn (and no 'start over' btn) is visible.
4. Failed draft form deletion via btn click on form page will display fail toast and will remain on current form page.

ps. Updated quite a few tests (lost the will to live doing them so not the prettiest looking things); will be amazed if e2e pass first time 😅
  

TIS21-4394 
TIS21-4877


components/common/PanelsCreator.tsx
```diff
@@ -64,7 +64,7 @@ export function PanelsCreator({
         <Card className={style.panelDiv}>
           <BodyText
             data-cy={`notAssigned${panelsName}`}
-          >{`You are not assigned to any ${panelsTitle}.`}</BodyText>
+          >{`You are not assigned to any p${panelsTitle.slice(1)}.`}</BodyText>
```
DorisWongNHS, line 67: Is the `p` in front of `${panelsTitle.slice(1)}` not added intentionally?

john12321: @DorisWongNHS, it is intentional; inconsistent 'P' flagged-up so 'un-capitalised' it.

DorisWongNHS (approved):

PR #934, opened 2023-08-23
chore: deployment pipeline details

ReubenRobertsHEE:
For the FE n00bs :)

NO-TICKET

### DIFF
--- a/cypress/component/forms/formr-part-a/FormA.cy.tsx
+++ b/cypress/component/forms/formr-part-a/FormA.cy.tsx
@@ -336,6 +336,14 @@ describe("FormA (form creation)", () => {
       .find(".react-select__option")
       .first()
       .click();
+    cy.get(
+      '[data-cy="programmeMembershipType"] > .autocomplete-select > .react-select__control > .react-select__value-container > .react-select__input-container'
+    )
+      .click()
+      .get(".react-select__menu")
+      .find(".react-select__option")
+      .first()
+      .click();  
     cy.get(".nhsuk-error-summary").should("not.exist");
     cy.wait(2000);
     cy.get('[data-cy="autosaveStatusMsg"]')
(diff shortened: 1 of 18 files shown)


john12321:
Really needed to emphasise pt 2 in that last commit !

DorisWongNHS (approved):

Judge40 (dismissed):
I'm happy as is, but I'm wondering whether it might make sense to have a sign in utility function that wrapped the wait, visit and sign in steps. We'd only have the explanation in a single place then and would have some more flexibility if we ever needed to change from a 30 second token refresh.

john12321:
> I'm happy as is, but I'm wondering whether it might make sense to have a sign in utility function that wrapped the wait, visit and sign in steps. We'd only have the explanation in a single place then and would have some more flexibility if we ever needed to change from a 30 second token refresh.

Should have stayed away from this 'NO-TICKET' 🤣

models/ProfileToFormRPartAInitialValues.ts
```diff
@@ -53,7 +53,7 @@ export function ProfileToFormRPartAInitialValues(
     completionDate: programme?.programmeCompletionDate ?? null,
     trainingGrade: "",
     startDate: programme?.startDate ?? null,
-    programmeMembershipType: programme?.programmeMembershipType ?? "",
+    programmeMembershipType: "",
```
ReubenRobertsHEE, line 56: Perhaps this should really go into another PR - I noticed that the dropdown for new forms was being prepopulated with 'MILITARY'. On staging for some reason the dropdown appears blank which is why I missed it (though it does actually have the value if you proceed to review and submit the form!)

cypress/component/forms/formr-part-a/FormA.cy.tsx
```diff
@@ -336,6 +336,14 @@ describe("FormA (form creation)", () => {
       .find(".react-select__option")
       .first()
       .click();
+    cy.get(
```
ReubenRobertsHEE, line 339: Tweak to component test to match change to ProgrammeMembershipType default

Judge40 (approved):

PR #943, opened 2023-09-07
fix: placement subspecialty display logic

ReubenRobertsHEE:
Allow subspecialties to be shown if they exist, regardless of postAllowsSubspecialty.

TIS21-4983

### DIFF
--- a/cypress/component/placements/Placements.cy.tsx
+++ b/cypress/component/placements/Placements.cy.tsx
@@ -11,6 +11,7 @@ import {
   mockPlacementNoOtherSites,
   mockPlacementPartialOtherSites,
   mockPlacementSubSpecialtyPostAllows,
+  mockPlacemenSubSpecialtyPostNotAllows,
   mockPlacementNoSubSpecialtyPostAllows,
   mockPlacementNoSubSpecialtyPostNotAllows
 } from "../../../mock-data/trainee-profile";
@@ -220,7 +221,7 @@ describe("Placements with MFA set up", () => {
       .should("contain.text", "sub specialty");
   });
 
-  it("should display Placements Subspecialty None Provided when Post Allows", () => {
+  it("should display missing Placements Subspecialty None Provided when Post Allows", () => {
     const MockedPlacementsSuccess = () => {
       const dispatch = useAppDispatch();
       dispatch(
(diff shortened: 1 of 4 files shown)


DorisWongNHS (approved):

cypress/component/placements/Placements.cy.tsx
```diff
@@ -248,7 +249,35 @@ describe("Placements with MFA set up", () => {
       .should("contain.text", "None provided");
   });
 
-  it("should not display Placements Subspecialty if Post does not Allow", () => {
+  it("should not display missing Placements Subspecialty if Post does not Allow", () => {
```
Judge40, line 252: These descriptions are the wrong way around I think?

Judge40 (changes requested):

cypress/component/placements/Placements.cy.tsx
ReubenRobertsHEE: Sorry, maybe poorly phrased. The intention was to test that placements without a subspecialty (i.e. 'missing') would not have the 'Sub specialty' field displayed at all if postAllowsSubspecialty = false

ReubenRobertsHEE: So not show 'Sub specialty: None provided' if postAllowsSubspecialty = false, basically.

Judge40: Ah, got it 🥴

ReubenRobertsHEE: I think I snuck in the merge just before your comment 😁  - if you like I can make a PR to fix the wording if you think its worth clarifying

ReubenRobertsHEE: Oh sorry, ok, looking at the code it is wrong - let me fix 🤦

Judge40: Actually, I'm not sure I do 😅 

e.g.
Description: "should display Placements Subspecialty even if Post does not Allow"
Mock data: `mockPlacementNoSubSpecialtyPostNotAllows`
Assertion: `cy.get('[data-cy="subSpecialty0Val"]').should("not.exist");`

The mock and the assertion match, but I can't get my head around the description of it "should display placements subspecialty" then asserting `not.exist`

PR #962, opened 2023-09-26
From UTC to London (time zone) minus the fudge 

john12321:
With an eye on the main bit of FE work for the DSP timestamp ticket https://hee-tis.atlassian.net/browse/TIS21-5138, the aim of this ticket is to have a consistent way of displaying the London time zone date/time minus the need to hardcode the time zone part (we currently add 'GMT' to all form submission dates whatever the season which is a bit 💩).

**Proposed solution** 
- Create a function to convert the UTC date to the London time zone (BST/GMT) date..
- Use this function to display a form submission date (without the seconds in the 'time' part as is - but we may want to change this?), and an autosave timestamp (with the seconds included in the 'time' part as is).
- The function now checks if the given date falls within an “ambiguous” period (between 1am and 2am on the last Sunday in October - which technically occurs  twice due to the change from BST to GMT. ). If the date is ambiguous then 'GMT' is appended to the end of the date string (as is the case now via hardcoded 'GMT')

NO TICKET

### DIFF
--- a/utilities/DateUtilities.ts
+++ b/utilities/DateUtilities.ts
@@ -121,6 +121,15 @@ export class DateUtilities {
     return true;
   }
 
+  private static isAmbiguousOctoberDate(date: day.Dayjs): boolean {
+    const octoberDate = day(date).month(9).date(31);
+    const lastSundayInOctober = octoberDate.day(0);
+    return date.isBetween(
+      lastSundayInOctober.hour(1).minute(0),
+      lastSundayInOctober.hour(2).minute(0)
+    );
+  }
+
   public static ConvertToLondonTime(
     givenDate: DateType,
     includeSecs?: boolean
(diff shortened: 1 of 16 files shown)


utilities/DateUtilities.ts
```diff
@@ -117,17 +121,36 @@ export class DateUtilities {
...
-      timeZone: "Europe/London",
-      hour12: false,
-      hour: "2-digit",
-      minute: "2-digit",
-      second: "2-digit"
-    };
-    const gbDate = new Date().toLocaleDateString("en-GB");
-    const gbTime = new Date().toLocaleTimeString("en-GB", timeZoneOptions);
-    return `${gbDate} ${gbTime}`;
+  private static isAmbiguousOctoberDate(date: day.Dayjs): boolean {
```
ReubenRobertsHEE, line 124: Ah yes. Don't worry, Single/Double Summer Time (SDST) will definitely not sort this out either :trollface:

ReubenRobertsHEE (approved):

PR #976, opened 2023-11-06
refactor: reword covid question in section 7 of formR partB

EdwardBarclay:
TIS21-5294


Judge40 (approved):

john12321 (approved):
Sorry for the bottleneck to stop you updating the wording (local dev env playing up today so testing has been a nightmare). Think I might need to fix the fix lol.

PR #980, opened 2023-11-13
feat: Replace the CoJ sign link with big green button

john12321:
TIS21-5334


Judge40 (approved):
Late night @john12321?

PR #992, opened 2023-12-21
fix: fix failing e2e tests

EdwardBarclay:
chore bump version

no-ticket

### DIFF
--- a/cypress/e2e/formr-b/FormRB.spec.ts
+++ b/cypress/e2e/formr-b/FormRB.spec.ts
@@ -188,18 +188,11 @@ describe("Form R (Part B) - desktop", () => {
     cy.get("[data-cy=LinkToNextSection] > .nhsuk-pagination__page").click();
 
     // -------- COVID Section ------------------------------------------------
+    //should be disabled whilst covide declaration flag is disabled in trainee-forms
     if (isCovid) {
       cy.log("### COVID SECTION CHECK ###");
-      cy.get(".progress-step")
-        .eq(6)
-        .should("have.class", "progress-step-active");
-      cy.get("[data-cy=LinkToNextSection] > .nhsuk-pagination__page").click();
-      cy.get("#haveCovidDeclarations--error-message").should("exist");
-      cy.get(".progress-step")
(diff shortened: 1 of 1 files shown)


cypress/e2e/formr-b/FormRB.spec.ts
```diff
@@ -211,30 +204,30 @@ describe("Form R (Part B) - desktop", () => {
...
-      );
-      cy.get("[data-cy=educationSupervisorEmail]").should(
-        "contain.text",
-        "No supervisor email provided"
-      );
-    } else {
-      cy.get("[data-cy=sectionHeader7]").should("not.exist");
-      cy.get("[data-cy=BtnEditSection7]").should("not.exist");
-    }
+    // if (isCovid) {
```
john12321, line 207: @EdwardBarclay Not sure why this code is commented out. Doesn't the if/else statement do its job? What happened when you left this code uncommented and ran the test locally? I'm assuming isCovid is set to false for this test?

john12321 (changes requested):

cypress/e2e/formr-b/FormRB.spec.ts
```diff
@@ -188,18 +188,11 @@ describe("Form R (Part B) - desktop", () => {
     cy.get("[data-cy=LinkToNextSection] > .nhsuk-pagination__page").click();
 
     // -------- COVID Section ------------------------------------------------
+    //should be disabled whilst covide declaration flag is disabled in trainee-forms
     if (isCovid) {
```
john12321, line 192: @EdwardBarclay Not too sure what's going on here (been a long day!). Looking at line 78 in this test file:  isCovid = true . Shouldn't isCovid be false (it is originally set to false on line 6) and then the if/else statements etc. can be left as they were before the fixes? Happy to look at it properly tomorrow.

john12321 (changes requested):

john12321 (approved):
Did leave another comment earlier but was stale after your changes so removed it. LGTM now isCovid is false 🤞

PR #993, opened 2023-12-22
Fix incorrect behaviour for covid section display logic

EdwardBarclay:
update the logic around weather the COVID declaration section appears in submitted forms and prevent it form showing in the final view when submitting a new form

TIS21-5513

### DIFF
--- a/components/forms/formr-part-b/viewSections/View.tsx
+++ b/components/forms/formr-part-b/viewSections/View.tsx
@@ -80,7 +80,9 @@ const View = ({ canEdit, history }: IView) => {
         <ViewSection4 {...viewSectionProps} />
         <ViewSection5 {...viewSectionProps} />
         <ViewSection6 {...viewSectionProps} />
-        {isCovidDeclarationNull ? <ViewSection7 {...viewSectionProps} /> : null}
+        {!canEdit && !isCovidDeclarationNull ? (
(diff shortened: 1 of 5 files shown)


components/forms/formr-part-b/viewSections/View.tsx
```diff
@@ -80,7 +80,7 @@ const View = ({ canEdit, history }: IView) => {
         <ViewSection4 {...viewSectionProps} />
         <ViewSection5 {...viewSectionProps} />
         <ViewSection6 {...viewSectionProps} />
-        {isCovidDeclarationNull ? <ViewSection7 {...viewSectionProps} /> : null}
+        {!canEdit && !isCovidDeclarationNull ? <ViewSection7 {...viewSectionProps} /> : null}
```
john12321: It might be worth making things a bit clearer here.. You could swap the logic round a bit so you are not having to negate a isNull variable . Something like renaming the `... isCovidDeclarationNull ... `to `const covidDeclarations = formData.haveCovidDeclarations` (on line 32). The Types for this variable will be `string, boolean, null `so this will still work to render/not render the covid section.
Then on line 83 you can do something like `{!canEdit && covidDeclarations && (<ViewSection7 {...viewSectionProps} />)}`

john12321:
@EdwardBarclay Will catch-up next week to tidy up commits etc. 🎄

PR #999, opened 2024-01-11
fix: the 'Hawaii problem'

john12321:
...where the user submits a form in e.g. Hawaii (UTC -10:00) timezone at say 10pm on 10/01/2024, which BE saves as 8am 11/01/2024 UTC. The result is a latest submitted 'future' form not being displayed in the Action Summary. Easiest fix is to remove the future constraint from the utils function logic (not sure why it was there in the first place tbh).

Also, after carefully considering the 'Doogie Howser, M.D. factor' with my learned colleague, R. Roberts esq. ,make the legalAge text match the D.O.B. validation logic (18+) - which in turn matches the BE validation logic.

NO TICKET


DorisWongNHS (approved):
It makes sense to me.

PR #1000, opened 2024-01-12
chore: remove redundant WTE .nullable() validation

john12321:
...from WTE validation schema. This might be a legacy thing but can't see the WTE type ever being null. Not saying this fixes the WTE issue but might narrow the focus.

NO TICKET


Judge40 (approved):
Happy 1000 🎉

PR #1004, opened 2024-01-30
Fix  date configuration on text input field date picker

EdwardBarclay:
The year section of the date picker was allowing an excess of numbers due to the layout of the date being yyyy/dd/mm instead of yyyy/mm/dd

When trying to create tests to check if more than 4 digits could be entered into the field, it was found that cypress was unable to enter the values into the date picker in such a way that would allow individual digits to be added. 

TIS21-5645

### DIFF
--- a/components/forms/TextInputField.tsx
+++ b/components/forms/TextInputField.tsx
@@ -49,7 +49,7 @@ const TextInputField: FunctionComponent<Props> = props => {
         {...rest}
         readOnly={props.readOnly}
         min="1920-01-01"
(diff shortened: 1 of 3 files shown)


john12321 (approved):
Goodbye old friend 😥

Pity about that cypress test stuff. Might be worth using https://www.browserstack.com/users/sign_in when merged to stage to quickly test on a few different devices/OS to see how it behaves. Probably a useful exercise in itself to see the native date input in action; you might spot some other weirdness?

PR #1006, opened 2024-02-02
ajust length of number fields to 3 allowing 5 numbers. add clearer la…

EdwardBarclay:
…beling to the TOOT fields

TIS21-5649


Judge40 (dismissed):

Judge40 (changes requested):
Changes LGTM, but you need to bump the version

PR #1012, opened 2024-02-14
Refactor FormBuilder - part 2

john12321:
Another WIP commit, the main objective:
Similar to #1009,  make sure Form R A still works while adding the FormBuilder logic to allow for number, total and textarea fields which are needed to build Form B.

Tests to follow when we build Form B using FormBuilder - which is probably the next step.
 
Note: Built separate Text and TextArea comps for now but can merge them into one Text component if it makes more sense.
 
TIS21-4591
TIS21-5718


Judge40 (approved):
Figured I'd get in before it grew 😂

PR #1013, opened 2024-02-15
Feat/programme review actions

DorisWongNHS:
TIS21-5582
TIS21-5618


john12321 (approved):
LGTM @DorisWongNHS . Nice work wrestling with the state management! 😁  (Looks good for this ticket but may need revisiting if/when the trainee actions logic gets more complex).

Will leave the more detailed discussions around button colour/width etc. to the others 🤣 

(ps. Re was about to mention the unused import but you beat me to it!).

PR #1014, opened 2024-02-16
feat: Add ScrollToTop to FormBuilder

john12321:
The purpose of this commit is to provide consistent behaviour for all forms by adding the ScrollToTop to the FormBuilder. 
(The scrollToTop used to be on the old Form A and there is one on Form B.)

TIS21-5729


DorisWongNHS (approved):
Looks good to me on local

PR #1018, opened 2024-02-23
feat: readme for the form builder

ReubenRobertsHEE:
Not currently implemented using typedoc.

Definitely needs review, preferably from someone who _knows_ :) . I got a bit lost with some of these fields, especially where the implementation within the form builder context seemed incomplete (e.g. if they were only applicable to FormR PartB).

TIS21-5641


ReubenRobertsHEE:
@john12321 thanks for the positive feedback. I've made a couple of minor tweaks, but as we said, this will be a WIP mirroring the FormR PartB work.

john12321 (approved):
Nice work, @ReubenRobertsHEE 🌮 . Looking forward to the sequel.📚

PR #1019, opened 2024-02-26
Feat/action due date

DorisWongNHS:
TIS21-5582
TIS21-5741

### DIFF
--- a/components/common/PanelsCreator.tsx
+++ b/components/common/PanelsCreator.tsx
@@ -91,7 +91,18 @@ export function PanelsCreator({
                         </Label>
                       </SummaryList.Key>
                       <SummaryList.Value>
+                        <p
+                          data-cy={`actionDueDate-${panelsName}-${panel.tisId}`}
+                        >
+                          Due by{" "}
+                          {DateUtilities.ToLocalDate(currentAction[0].dueBy)}{" "}
+                          {today >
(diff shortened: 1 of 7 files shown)


Judge40 (changes requested):
Is the overdue marker something you've discussed with Naz?

We don't have any ACs for this and not all actions will actually become overdue so more planning is needed for how that will be displayed.

DorisWongNHS:
@Judge40 Naz hasn't came back to me yet. But as it will need further discussion. I have removed the `overdue` label in this PR to make it simple. It can be added with separated ticket or PR later.

DorisWongNHS:
As Naz has confirmed to remove the `overdue` label. Changes have been pushed to the branch, and it is ready to be reviewed again.

john12321:
> _ No description provided. _
Hi @DorisWongNHS, A PR description/ tickets numbers would be good. (At first, wasn't sure "Show localise date on FE" was part of this work. Easily confused/ still half asleep coming back from leave 😀).
If you still think it's a patch bump then you'll need to bump to 3 (Reuben has already claimed 2 😄 ).

DorisWongNHS:
Thank you @john12321 it has been rebased and bump the patch version 3 now 😄

ReubenRobertsHEE (approved):
Looks good.

PR #1022, opened 2024-02-27
Add submission declarations checklist to FormBuilder 

john12321:
**Problem**
There is currently no consistency with Form R submission declarations.
Form A currently has a warning callout with a "by submitting this form..." declaration, with no mention of any agreements on the submitted form View page, while Form B does have checkboxes to tick as part of the main form.

**Solution** 
1. Standardise the submission checklist so the checklist fields and text can be passed in via the JSON file to a self-contained Declarations component.
2. Simplify the checkbox validation logic i.e. ‘Submit Form’ button is disabled unless all checkboxes have been checked (with no need for additional form validation).

Note: Some of the submit logic added to `FormAView.tsx` has been copied from the existing logic in `formr-part-b/sections/Declarations.tsx` .

TO DO:
There's still scope to refactor this code further but the main priority is to have something that can be used for a form built via the FormBuilder.
A lot of the other Warning text can be moved to the JSON file and passed into the form in a similar way to the Declarations.

TIS21-5754

### DIFF
--- a/components/forms/form-builder/FormBuilder.tsx
+++ b/components/forms/form-builder/FormBuilder.tsx
@@ -77,6 +77,7 @@ type Section = {
 export type Form = {
   name: string;
   pages: Page[];
+  declarations: { name: string; label: string }[];
 };
 type FormBuilderProps = {
   jsonForm: Form;
--- a/components/forms/form-builder/Declarations.tsx
+++ b/components/forms/form-builder/Declarations.tsx
@@ -1,0 +1,53 @@
+import { Checkboxes } from "nhsuk-react-components";
+import { Form } from "./FormBuilder";
+import { useEffect, useMemo, useState } from "react";
+
+type DeclarationsProps = {
+  setCanSubmit: (canSubmit: boolean) => void;
+  canEdit: boolean;
+  formJson: Form;
+};
+
+export default function Declarations({
+  setCanSubmit,
+  canEdit,
+  formJson
+}: Readonly<DeclarationsProps>) {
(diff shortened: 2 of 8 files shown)


components/forms/form-builder/FormBuilder.tsx
```diff
@@ -77,6 +77,7 @@ type Section = {
 export type Form = {
   name: string;
   pages: Page[];
+  declarations: { name: string; label: string }[];
```
ReubenRobertsHEE, line 80: You don't think making this a proper type is worth doing at this stage? (Also, any inclination to update that lovely README 😈 ?)

ReubenRobertsHEE:
I like this a lot, btw.

components/forms/form-builder/FormBuilder.tsx
john12321: @ReubenRobertsHEE Thanks for reviewing on a Fri pm 💪.
RE the type, the whole `declarations` prop is a bit of a WIP at the moment; was umming and ahhing on whether or not to make this into something that could handle more than just the submission declarations but then pictured your wagging finger...A similar excuse for the README update too and I'm sticking to it 😁 .
Happy to push some changes with this commit though if it helps to get this one over the line.

ReubenRobertsHEE: Good point: I should appreciate these wafer-thin PRs for what they are! 😍

ReubenRobertsHEE (approved):

PR #1026, opened 2024-03-21
Add in-app notifications (welcome pt1)

john12321:
Add Notifications list, message page, icon

@ReubenRobertsHEE AKA last-man-standing...

I've copied these notes I made on the second commit which might be useful to find your way around the chnages:

1. Notifications.tsx - routing entry point for notifications page
2. NotificationsTable.tsx - /notifications to list the notifications. This component houses some child components: DebouncedInput, AllUnreadCheckbox, TablePagination. It also has the click event logic to mark a row as read/navigate to open the message (line 105). The autoResetPageIndex prop is set to false so when a paginated row is clicked the table doesn't reset.
3. columns.tsx is where the table columns are defined. It houses the TableColumnHeader and RowActions components.
4. RowActions.tsx - deals with the 'mark as unread' btn click event logic.
5. NotificatiosUtilities.ts - This has a updateNotificationStatus function that is called from the NotificationsTable (row click) and RowActions (mark as unread btn click). Still a bit of a WIP. Current implementation is to update the notification status optimistically and then call the API to update the status. If the API call fails, the status is reverted back to the original status (you can see this best with slow internet!).
6. NotificationMessage.tsx - houses the NotificationMessageText component and dispatches the action to get the message from the backend on page load using the id in the url. There's a bit of error handling (using the ErrorPage component) if the message is not found etc.
7. TablePgination.tsx - page sizes are hardcoded for now. Mobile view needs fixing (Ed is looking at this - could be a follow-up commit?).
8. StringUtilities.ts - added a function to truncate subject text of your choosing. Wasn't sure on size but we can play around with it.

Tested as much as I can of the table interactions etc. but couldn't test the complete journey from e.g. marking a notification as unread -> optimistic update -> fail -> revert optimistic changes using Cypress Component. Got a bit further with RTL but would involve a lot more mocking with diminishing utility. Maybe e2e is the way to go with this? Speaking of which, there's just a really basic notifications e2e given the fragile nature of our test data 💥.  

Please feel free to change any of the logic etc. if you spot a better solution. I found the react-table docs https://tanstack.com/table/latest pretty useful.

TIS21-5682

### DIFF
--- a/components/common/ErrorPage.tsx
+++ b/components/common/ErrorPage.tsx
@@ -1,0 +1,28 @@
+type ErrorPageProps = {
+  message: string;
+  header?: string;
+};
+
+export default function ErrorPage({
+  message,
+  header = "Oops! Something went wrong"
+}: Readonly<ErrorPageProps>) {
+  return (
+    <div
+      className="nhsuk-error-summary"
+      aria-labelledby="error-summary-title"
+      role="alert"
+    >
+      <h2
+        className="nhsuk-error-summary__title"
+        id="error-summary-title"
+        data-cy="error-header-text"
+      >
+        {header}
+      </h2>
+      <div className="nhsuk-error-summary__body" data-cy="error-message-text">
+        {message}
+      </div>
+    </div>
+  );
+}
--- a/components/main/Main.tsx
+++ b/components/main/Main.tsx
@@ -30,6 +30,7 @@ import packageJson from "../../package.json";
 import { loadFormAList } from "../../redux/slices/formASlice";
 import { loadFormBList } from "../../redux/slices/formBSlice";
 import { fetchTraineeActionsData } from "../../redux/slices/traineeActionsSlice";
+import { Notifications } from "../notifications/Notifications";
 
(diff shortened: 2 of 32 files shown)


ReubenRobertsHEE:
So far I like it a lot. I've tested everything I can think of (pagination, broken notifications service, reading/unreading, search, sorting). All looks good. If I had to have a quibble, its the placement of the button slightly below the other menu items 🔬 
image

ReubenRobertsHEE:
Yeah, mobile view could use a bit of work. 
image

ReubenRobertsHEE:
One tiny glitch: you can't enter a page greater than the maximum (e.g. if you type '9' when the biggest page is 2, it turns into a '2'). But you _can_ happily type -235235235 - it doesn't break anything though 😁

ReubenRobertsHEE (dismissed):
A couple of things we should probably pick-up in a future PR (esp. mobile layout) - but it looks really good: let's get Naz onto it! 😀

ReubenRobertsHEE (approved):
👍

PR #1027, opened 2024-03-22
fix: navbar e2e test

ReubenRobertsHEE:
The current styling markup is a little clumsy, but it is scheduled for revision, so I haven't done more than patch the e2e test error.

TIS21-5684

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.94.1",
+  "version": "0.94.0",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.94.0",
+  "version": "0.94.1",
```
hihilary, line 3: Just a tiny question as I have no idea at all. Do we bump up version for test fix?

ReubenRobertsHEE: Good point. I'll actually revert that.

hihilary (approved):

PR #1029, opened 2024-03-27
Filter by notification type

EdwardBarclay:
fix to filter the received notifications so that only those marked type "IN_APP" are displayed to the user to avoid being spammed by email notifications

TIS21-5685


Judge40 (changes requested):
Just missing a version increment (patch I guess)

Judge40 (approved):

PR #1032, opened 2024-04-12
fix: remove 13-week CoJ signing horizon

ReubenRobertsHEE:
Note, as discussed, the new-starter logic will be implemented separately as TIS21-5914

TIS21-5890
TIS21-5739

### DIFF
--- a/cypress/component/programmes/ConditionsOfJoining.cy.tsx
+++ b/cypress/component/programmes/ConditionsOfJoining.cy.tsx
@@ -136,31 +136,7 @@ describe("ConditionsOfJoining", () => {
         .and("have.text", "Follow Local Office process");
     });
 
-    it("should show available signing date when start date after COJ epoch but outside 13 weeks", () => {
-      mount(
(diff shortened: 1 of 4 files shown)


john12321:
> As a side-issue, do we need to reset our SonarCloud analysis for the component somehow? It seems to pick up issues in lots of code untouched by the PR. And/or do we need to attend to the recommendations? 😬

Was wondering if sonarcloud just gets its own code updated and feels reinvigorated to have another pass over old stuff too? Probably not a bad idea if true.
The [failure](https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=BUG&pullRequest=1032&id=Health-Education-England_trainee-ui&open=AY7SnOQWUZHt7bUzRZwU) is a dead simple fix - which could done on the next PR (promise 😁).  

See you've merged it anyway. That's the spirit, lad!

PR #1033, opened 2024-04-23
feat: add static assets to linking from email

ReubenRobertsHEE:
There may be a better place for these @john12321 🙈 

Included so that emails can link images via full URL.

TIS21-5928


ReubenRobertsHEE:
Context: https://hee-nhs-tis.slack.com/archives/C03G3UK3U82/p1713881503279079

PR #1040, opened 2024-05-10
feat: Add Placement Other Specialties

john12321:
TIS21-5744


ReubenRobertsHEE (commented):
Looks ace to me, thanks John! Let's check with Naz how she wants to handle this, since there is a problem with the completeness of the data in the sync service, which might not get resolved for some days.

components/common/PanelsCreator.tsx
```diff
@@ -214,6 +214,16 @@ function displayTheCorrectListItem(
...
+    case "otherSpecialties":
+      return panel[panelProp]?.length > 0
+        ? panel[panelProp].map(
+            ({ specialtyId, name }: SpecialtyType): JSX.Element => (
+              <div key={specialtyId}>
+                <div data-cy={`otherSpecialty${specialtyId}Val`}>{name}</div>
+              </div>
+            )
+          )
+        : "None provided";
```
ReubenRobertsHEE, line 226: One thing I was wondering: less than 1% of placements have 'Other specialties' - is it worth using up screen real estate to show 'None provided'? Or is it just easier to keep it consistent with how all of the placement details are displayed?

john12321: Yeah, bit of a judgement call this one. Real estate-wise makes little difference and as you say, it’s consistent. Also showing it might trigger those users who expect to see something there etc.

john12321:
Will re-open when Other specialties data sync blocker is resolved.

PR #1041, opened 2024-05-12
Feat : Add Form R(B) via FormBuilder

john12321:
The first commit 8e44e0853b68f7da10a6eb3328ee6b07cf51bad3 is where most of the implementation action is.
As far as the new layout for FormB goes, no massive changes just an effort to move towards a more GDS-friendly approach by having a single section per page. 

Some follow-up stuff to consider:
When editing the form, the 'shortcut' button appears on all the form pages; it probably shouldn't appear on the last page (as is the case with the current Form B implementation) given that the 'next' navigation gets you there just as quick!
Move more of the text into the JSON file (e.g. important text)
Make the validation schemas a bit more DRY.
Lost the will to live with the test files but they could be more DRY too.

TIS21-5979


components/forms/form-builder/form-r/part-b/formBValidationSchema.ts
```diff
@@ -0,0 +1,257 @@
...
+    .test("endDate", "The date is outside the allowed date range", value =>
+      DateUtilities.IsInsideDateRange(value)
+    )
+});
+
+const covid19ValidationSchema = yup.object().shape({
+  selfRateForCovid: yup
+    .string()
+    .nullable()
+    .required("Self-rating your own preformance is required"),
```
Judge40, line 97: Typo in "performance"

redux/slices/formBSlice.ts
```diff
@@ -53,10 +70,14 @@ export const loadFormBList = createAsyncThunk(
 export const loadSavedFormB = createAsyncThunk(
   "formB/fetchFormB",
   async (formId: string) => {
+    const state = store.getState();
+    // Grab the Covid flag status from the feature flags slice to use to deterimine if the Covid section should be displayed and if a default Covid object should be added to the form data
```
Judge40, line 74: Typo in "determine"

Judge40 (changes requested):
I spotted a couple of minor typos, but otherwise LGTM 😵

john12321:
> I spotted a couple of minor typos, but otherwise LGTM 😵

Good spots! Powered by 'Spoons? 🍻

redux/slices/formBSlice.ts
john12321: I quite like deterimine. Defo belongs in the Covid section.

Judge40 (approved):

PR #1051, opened 2024-05-23
chore: fix sonar analysis

Judge40:
Sonar analysis is failing as the repo is not checked out. Perform a checkout with deep clone before attempting to run sonar.

NO-TICKET


john12321 (approved):
Your FE metrics are off the chart! 🤣

PR #1052, opened 2024-05-24
Reword dsp ui buttons

EdwardBarclay:
TIS21-6064


john12321:
Hi Ed, given the changes per commit would look to squash them if that's OK?

PR #1055, opened 2024-05-24
fix: e2e wait time increased

john12321:
...between form b tests to allow MFA refresh
(should have put it to 30s in the first place 😂)

NO TICKET


ReubenRobertsHEE (approved):
But John, why so few lines, has it been a long week mate? 😁

PR #1060, opened 2024-06-07
fix: alphabeticise other specialties

ReubenRobertsHEE:
@john12321 - look ok to you? Feels a bit clumsy, but that's 'fullstackers' for you 😁

### DIFF
--- a/components/common/PanelsCreator.tsx
+++ b/components/common/PanelsCreator.tsx
@@ -215,15 +215,20 @@ function displayTheCorrectListItem(
         />
       );
     case "otherSpecialties":
-      return panel[panelProp]?.length > 0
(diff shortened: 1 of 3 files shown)


components/common/PanelsCreator.tsx
```diff
@@ -215,15 +215,20 @@ function displayTheCorrectListItem(
         />
       );
     case "otherSpecialties":
-      return panel[panelProp]?.length > 0
-        ? panel[panelProp].map(
+      if (panel[panelProp]?.length > 0) {
+        var otherSpecialties = [...panel[panelProp]];
```
john12321, line 219: `var `? You maverick! 😀

PR #1063, opened 2024-06-11
feat: replace currRevalDate hard validation

john12321:
...with soft validation warning msg.

Was a bit more fiddly than expected because the function to display warning was expecting a regex. 

See [slack thread](https://hee-nhs-tis.slack.com/archives/G0135LS4JVA/p1718104244591819) for further details.

NO TICKET


ReubenRobertsHEE (approved):
🤞

PR #1066, opened 2024-06-12
Feat: Add CCT Calculator to TSS

john12321:
Add basic CCT form to calculate new end dates for single or multiple WTE changes (higher to lower WTE(s) and vice versa).
Allow user to add bespoke values for WTE percentages.
Add some agreed defaults and basic form validation.
Add some warnings for user to consider (e.g. change date < 16 weeks, bespoke WTE).
Allow CCT calc to remain active as user interacts with other parts of the app.
Fix CCT position on phone/small mobile to top of page; Make CCT draggable on other devices.
Add 'further info' as expander.

Set print preview for draggable dialog.
Target dialog element so only this shows in print preview (reset its position before preview so that it appears in the correct position for printing).

Note:
Test CCT calc functionality via browserstack using multiple devices before pushing to prod.
Hotjar survey will be embedded on TSS Programmes page to get end-user feedback to help us iterate on this MVP.

### DIFF
--- a/cypress/component/forms/cct/Cct.cy.tsx
+++ b/cypress/component/forms/cct/Cct.cy.tsx
@@ -1,0 +1,161 @@
+/// <reference types="cypress" />
+/// <reference path="../../../../cypress/support/index.d.ts" />
+import { mount } from "cypress/react18";
+import { MemoryRouter } from "react-router-dom";
+import { Provider } from "react-redux";
+import store from "../../../../redux/store/store";
+import { Cct } from "../../../../components/forms/cct/Cct";
+import {
+  openCctModal,
+  setCurrentProgEndDate,
+  setNewEndDates,
+  setProgName
+} from "../../../../redux/slices/cctCalcSlice";
+import dayjs from "dayjs";
+
+describe("Cct", () => {
+  it("renders the Cct component", () => {
+    const progName = "General Practice";
+    const endDate = dayjs().add(3, "year");
+    store.dispatch(setNewEndDates([]));
+    store.dispatch(openCctModal());
+    store.dispatch(setProgName(progName));
+    store.dispatch(setCurrentProgEndDate(endDate.format("YYYY-MM-DD")));
+
+    const newEndDate = (newPercent: number, currPercent: number) => {
+      const propStartDate = dayjs().add(16, "week").subtract(1, "day");
+      const chunkDays = endDate.diff(propStartDate, "days");
+      const chunkDaysWTE = Math.ceil((chunkDays * currPercent) / newPercent);
+      return dayjs(endDate)
+        .add(chunkDaysWTE - chunkDays, "days")
+        .format("DD/MM/YYYY");
+    };
+    mount(
+      <Provider store={store}>
+        <MemoryRouter initialEntries={["/programme"]}>
+          <Cct />
+        </MemoryRouter>
+      </Provider>
+    );
+    cy.get('[data-cy="cct-header"]').contains("CCT Calculator");
+    cy.get('[data-cy="cct-disclaimer"]').contains(
+      "This calculator is intended to provide a quick rough estimate only. Your actual CCT date will be confirmed at your next ARCP."
+    );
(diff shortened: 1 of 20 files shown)


ReubenRobertsHEE:
+3,386 −2,381 
😂

john12321:
I promise to be net zero by 2030!

ReubenRobertsHEE:
Ah, its mainly the package-lock.json changes. You are perhaps forgiven this once. ✝️

john12321:
Bless you, Father. ☦

ReubenRobertsHEE:
Well, the only way I found to get it to produce funny output was to give it a funny programme 😁 
image

ReubenRobertsHEE:
Given there are a range of browser dimensions where the calculator obscures the programme, would it be worth including the programme start date in the description on the calculator, so that all the pertinent details are available there?

ReubenRobertsHEE:
Would it be worth stopping someone entering a custom WTE that has more than a few digits after the decimal point? Or 'technically correct' values like 1e2? It doesn't actually break the calculator, but it don't look pretty.

ReubenRobertsHEE:
Another slight funny where the 'When should the WTE change start' is before the programme start date:
image

cypress/component/forms/cct/Cct.cy.tsx
```diff
@@ -0,0 +1,161 @@
...
+  it("renders the Cct component", () => {
+    const progName = "General Practice";
+    const endDate = dayjs().add(3, "year");
+    store.dispatch(setNewEndDates([]));
+    store.dispatch(openCctModal());
+    store.dispatch(setProgName(progName));
+    store.dispatch(setCurrentProgEndDate(endDate.format("YYYY-MM-DD")));
+
+    const newEndDate = (newPercent: number, currPercent: number) => {
+      const propStartDate = dayjs().add(16, "week").subtract(1, "day");
```
ReubenRobertsHEE, line 26: Maybe this should default to the programme start date if this is more than 16 weeks in the future?

redux/slices/cctCalcSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+  newEndDates: any[];
+  propStartDate: Date | string;
+};
+
+const initialState: CctCalcState = {
+  modalOpen: false,
+  progName: "",
+  currentProgEndDate: "",
+  newEndDates: [],
+  propStartDate: dayjs().add(16, "weeks").format("YYYY-MM-DD")
```
ReubenRobertsHEE, line 17: Maybe this should default to the programme start date if this is more than 16 weeks in the future? (And not be allowed to be before the programme start date)

john12321:
> Well, the only way I found to get it to produce funny output was to give it a funny programme 😁 (it seems to consider the 'When should the WTE change begin' even if this is before the programme start date)
> 
> image
> 
> 

Fair comment, yes there a few ways to break it but do you think this is a dealbreaker? Happy to spend more time polishing the MVP ;)

john12321:
Thanks again for diving in to review this! There's a common theme to your comments. Do any/all things need working on now or do you think most users will have "sensible" values that will give them what they want?

ReubenRobertsHEE:
I think the only thing might be the default and rule for 'When should the WTE change start' for future-dated programmes. But if Naz is happy then I am happy :)

ReubenRobertsHEE:
I'm going to take the 🐕 for a 🚶 so I'll approve as-is; if you decide to tweak I'll re-review a little later (if no-one else has jumped on it). I get the MVP aspect though.

john12321:
Cheers! And thanks again for the feedback. I will look at how we can implement some of your suggestions; will probably look at a follow-up commit for any quick wins. Enjoy the walk! Think you're the last FS dev standing mate! ;)

ReubenRobertsHEE (approved):
Sorry, forgot to approve before leaving! 🤦

john12321:
2 Hail Mary's should cover it!

PR #1069, opened 2024-06-18
feat: polish the CCT MVP(1)

john12321:
feat: update cct btn wording to better reflect the action
chore: add calcDefaultPropStartDate util (mainly to address the '>16 weeks future programme' issue).
chore: set the draggable cct dialog boundary to parent component to stop it getting completely hidden and unclickable.
chore: make the expander message text less wordy.
chore: add line breaks to disclaimer text (to stop some Safari weirdness with dialog width/print preview (to save PDF)

TIS21-6118

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.99.0",
+  "version": "0.100.0",
   "private": true,
   "dependencies": {
     "@aws-amplify/ui-react": "^5.3.0",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "trainee-ui-app",
-  "version": "0.99.0",
+  "version": "0.100.0",
```
Judge40, line 3: 🎉 100 minors versions without a a full release 🙃

john12321: What a milestone; here's to the next 100 ! 😆

PR #1083, opened 2024-08-02
Link FormR A & B submission to a programme membership/ ARCP

john12321:
feat: Add FormR linker logic to meet [ticket AC's](https://hee-tis.atlassian.net/browse/TIS21-5611)

TODO: 
- Follow-up refactoring around the forms-related Types and functions in a follow-up commit.
- Some accessibility changes needed e.g.  label's for attribute doesn't match any element id on some inputs.
- Would be nice for CCT Calc to use the new Modal comp.

### DIFF
--- a/components/forms/cct/CalcForm.tsx
+++ b/components/forms/cct/CalcForm.tsx
@@ -23,7 +23,7 @@ export type CalcFormValues = {
   ftePercents: FtePercentsTypes[];
   propStartDate: string;
   propEndDate: string;
-  currentFtePercent: string;
+  currentFtePercent: number | string;
 };
 
 export function CalcForm({
@@ -86,7 +86,7 @@ export function CalcForm({
   return (
     <Formik
       initialValues={{
-        currentFtePercent: "100%",
+        currentFtePercent: 100,
         ftePercents: [],
         propStartDate: propStartDate,
         propEndDate: currentProgEndDate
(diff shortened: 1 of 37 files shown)


components/forms/cct/CalcForm.tsx
```diff
@@ -23,7 +23,7 @@ export type CalcFormValues = {
   ftePercents: FtePercentsTypes[];
   propStartDate: string;
   propEndDate: string;
-  currentFtePercent: string;
+  currentFtePercent: number | string;
```
ReubenRobertsHEE, line 26: just had to slip this tweak in, eh? 😂

ReubenRobertsHEE (approved):
The rules for what PM is available for 'new starter' linking seem unusually liberal, but I accept they were agreed on.

Sadly, I couldn't break anything, and all the A/Cs seem to work just fine 😢

components/forms/cct/CalcForm.tsx
john12321: All killer no filler, my friend.
There was an error in the `AutocompleteSelect handleChange` function (incorrectly targeting `val.label` instead of the `val.value`) which then developed into a game of splat the rat. The rat may have won.

ReubenRobertsHEE: 🐀 🏅 😂

PR #1087, opened 2024-08-07
fix: the fix to the power of fix

john12321:
Targeting the btn element didn't work so instead add some 'wait' cmds before the btn click to ensure it is loaded/ state is reset.

NO TICKET


Judge40 (approved):
Approved<sup>Approved</sup>

PR #1090, opened 2024-08-22
refactor: rename local offices

Judge40:
The local offices are being renamed to remove the "Health Education England" prefix.
As there is no clean switch-over point both the new and old names need to be supported temporarily.

Update the contact list to include the new local office names alongside the existing HEE based LO names.

Update mock data and test expectations to use the new local office names.

TIS21-6418
TIS21-6419


john12321 (approved):
LGTM

PR #1091, opened 2024-08-23
fix: highlight actionable panels

ReubenRobertsHEE:
TIS21-6342


john12321 (approved):
Not as good as the magic eye but beggars can't be choosers.

PR #1094, opened 2024-09-02
feat: COJ Gold Guide 10 with separate COJ view

DorisWongNHS:
This is the downloaded GG10 PDF from the local FE for your reference.
[COJ_GG10.pdf](https://github.com/user-attachments/files/16873146/COJ_GG10.pdf)

TIS21-6426
TIS21-6434


components/forms/conditionOfJoining/CojView.tsx
```diff
@@ -25,21 +29,128 @@ export default function CojView() {
...
+      )}
+      {signingCojVersion === "GG10" && (
+        <CojDeclarationSection10 signedDate={signedDate} />
+      )}
     </>
   ) : null;
 }
 
-function CojDeclarationSection({
+function CojDeclarationSection9({
```
john12321, line 53: Hi @DorisWongNHS , to reduce the code duplication I would focus on making a more generic CojDeclarationSection component rather than 2 separate CojDeclarationSection9/10 components. You can just pass in props for any differences e.g. initialValues, validationSchema (where the only difference is `isDeclareContacted` ?), COJ_DECLARATIONS...

DorisWongNHS: Thank you JohnO, I will have a try!

john12321: Great stuff! A possible implementation for the yup validation schema object prop is in the `formAValidationSchema.ts`  (scroll to line 101) where you can have a default schema object then have any additional properties e.g. isDeclareContacted as the rest.

PR #1095, opened 2024-09-03
Add artist's impression of an onboarding tracker

john12321:
Add a new fully-immersive onboarding journey diagram for the ultimate mobile and desktop 'new starter' experience.
Allow user to: 
a. click relevant links on journey cards to navigate to the correct section to complete a given task,
b. click on info icon in a given action card, to open a modal displaying further details/links,
c. click outside modal to close it (in addition to already being able to close via 'close' btn, escape key).

Note:
1. Initially developed this thing assuming active statuses as per ACs so there is quite a bit of redundant action status-related logic lying round until Reuben's PR is ready - which will generate a few code smells in the meantime. Could refactor to remove in a follow-up commit if this is too confusing/messy. 
2. The CCT calc does not currently close via click outside as it was developed before the Modal comp so needs refactoring.

TIS21-6367

### DIFF
--- a/components/Common.module.scss
+++ b/components/Common.module.scss
@@ -37,6 +37,18 @@ $alertYellow: #ffeb3b;
   border: 10px solid $alertYellow;
 }
 
+.panelHeader {
+  font-size: 1.5rem;
+  font-weight: bold;
+}
+
+.panelSubHeader {
+  font-size: 1.25rem;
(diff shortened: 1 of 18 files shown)


john12321:
Will re-open when rebase etc. after @DorisWongNHS 's merge.

PR #1096, opened 2024-09-03
Add artist's impression of an onboarding tracker

john12321:
Add a new fully-immersive onboarding journey diagram for the ultimate mobile and desktop 'new starter' experience.

Allow user to:
a. click relevant links on journey cards to navigate to the correct section to complete a given task,
b. click on info icon in a given action card, to open a modal displaying further details/links,
c. click outside modal to close it (in addition to already being able to close via 'close' btn, escape key).

(Hopefully) fix eslint error on Analyse PR action run (by assigning modalRef.current to variable inside handleClickOutside useEffect to fix its value for the return clean-up). 

Note:
1. Initially developed this thing assuming active statuses as per ACs so there is quite a bit of redundant action status-related logic (until Reuben's PR is ready) which will probably generate a few code smells in the meantime. Could refactor to remove in a follow-up commit if this is too confusing/messy.
2. The CCT calc does not currently close via click outside as it was developed before the Modal comp so needs refactoring.

TIS21-6367

### DIFF
--- a/utilities/Constants.tsx
+++ b/utilities/Constants.tsx
@@ -1,4 +1,13 @@
+import {
+  faCircle,
+  faCircleQuestion,
+  faEnvelope,
+  faShare,
+  faUserFriends
+} from "@fortawesome/free-solid-svg-icons";
+
 import { Declaration, Work } from "../models/FormRPartB";
+import { Link } from "react-router-dom";
 
 export const CCT_DECLARATION =
   "I have been appointed to a programme leading to award of CCT";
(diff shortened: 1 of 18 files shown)


utilities/Constants.tsx
```diff
@@ -289,3 +298,208 @@ export const localOfficeContacts: localOfficeContactsProps = {
...
+    <p>
+      If you do notice any discrepancies when reviewing the{" "}
+      <Link to="/programmes">Programme</Link> data, please contact{" "}
+      <Link to="/support">Local Office support</Link>.
+    </p>
+  ),
+  SIGN_COJ: (
+    <>
+      <p>
+        You can sign/ view the signed Conditions of Joining Agreement via the
```
Judge40, line 430: Missing space in "sign / view"

utilities/Constants.tsx
```diff
@@ -289,3 +298,208 @@ export const localOfficeContacts: localOfficeContactsProps = {
...
+      The 'Placement Confirmation' email is sent to the email address you use to
+      sign in to TIS Self-Service.
+    </p>
+  ),
+  REVIEW_PLACEMENT: (
+    <>
+      <p>
+        When you receive the 'Placement Confirmation' email 12 weeks before your
+        start date, your Placement should be in{" "}
+        <Link to="/placements">Upcoming Placements</Link> for you to review.{" "}
```
Judge40, line 488: Extra whitespace at the end can be removed.

Judge40 (changes requested):
Just a couple of bits of text that need fixing, but otherwise LGTM

Judge40 (approved):

PR #1100, opened 2024-09-27
Feat/edit gmc2 johnny

ReubenRobertsHEE:
TIS21-6292


ReubenRobertsHEE:
@john12321 - is there anything else you would like covered test-wise? I'm guessing you can't review this (seeing as how much you've put into it), but any parting remarks or other concerns?

john12321:
> @john12321 - is there anything else you would like covered test-wise? I'm guessing you can't review this (seeing as how much you've put into it), but any parting remarks or other concerns?

@ReubenRobertsHEE Just seen your e2e test commit. You are the Cypress Sensai 🥋. Was thinking about adding a gmcStatus loading state test to the Profile.cy.tsx but you've covered it much more realistically in the e2e test.

Judge40:
Holy commit history Batman! 😂

cypress.config.ts
```diff
@@ -24,8 +24,8 @@ export default defineConfig({
     supportFile: "cypress/support/component.ts"
   },
   e2e: {
-    baseUrl: "https://stage.trainee.tis.nhs.uk/",
-    // baseUrl: "http://local.tis-selfservice.com",
+    //baseUrl: "https://stage.trainee.tis.nhs.uk/",
+    baseUrl: "http://local.tis-selfservice.com",
```
Judge40, line 28: I'm assuming this needs reverting

cypress/e2e/profile/Profile.spec.ts
Judge40, line 1: Is there any value in an additional test which explicitly verifies that an invalid value like `UNKNOWN` is retained if the user cancels the edit request, just to ensure we don't have any weird interactions with the validation?

Judge40 (changes requested):

cypress/e2e/profile/Profile.spec.ts
ReubenRobertsHEE: Sure, I'll push one (and revert that e2e config change that crept in)

Judge40 (approved):

PR #1115, opened 2024-12-13
chore: revert the test sign-in wait time

john12321:
... on Programmes.spec

NO TICKET


ReubenRobertsHEE (approved):
Dubious, but I'll let it pass.

PR #1138, opened 2025-02-11
Fix/formr submission bug 

john12321:
Couldn't isolate the exact cause of the bug where the same form was being both submitted and saved as a draft via the click event (!) but the fix is an attempt to reduce such unwanted side effects by simplifying the form updating/saving logic.

Some context:
The previous merge commit https://github.com/Health-Education-England/tis-trainee-ui/commit/3b04bbb51ce7e211efd4ae2d0ebe4565c661acba focused on separating concerns between the form rendering (via the FormBuilder)
 and the form state management (e.g. keep track of the formData changes made by the user completing the form).
The form state management is handled by the FormProvider and we update and access user-focused form state e.g. formData via the context hook (The form state was previously handled directly in the FormBuilder).
The saved form state which includes state not changed directly by the user (e.g. id) continues to be managed by the redux store.

1. Managing any form state divergence
The trickiest bit of the state management is managing the divergence between the FormProvider context formData and the redux store saved form data
 - in particular when the form is autosaved following a user change and the redux form state has an id but the context form state does not.
 This is easily handled by creating a newFormId state variable in the formA/B redux store to ensure subsequent autosaves are PUT requests. [see e.g. FormASlice.ts line 35]

2. Another tricky case is the handover between the FormBuilder formData state and the FormView [see continueToConfirm utils function].
The simplest approach I found was to continue to use the context formData as the source of truth (i.e. the autosave might not succeed between the last form page and confirm page) and use this to update the form's redux state.
This ensures that the last onChange form event is captured. With the id safely stored separately from the formData obj via newFormId, we are then able to add this in when needed.

3. Simplifying the form save/update/submit logic [see utilities/FormBuilderUtilities.ts]
Create A single saveFormR function with additional params for isAutoSave and isSubmit. is used to save/update the form data.
(The redux boilerplate is very repetitive and can be refactored later). It also handles the creation of the save/update/submit object via the prepFormData utils function.
 [see e.g. FormASlice.ts line 215] The isSubmit flag allows for the correct 'submit' toast message to be displayed.
 
 (NOTE: did look briefly at subscribing to the redux store directly in the FormContext via a useEffect hooked up to the form's store but it has unwanted re-rendering issues on each autosave. Who's idea was autosave anyway!? 😅)
 
 Hope this makes sense. Happy to explain more/ run through the code changes.


DorisWongNHS (approved):
LGTM

PR #1141, opened 2025-02-17
refactor: ltft summary table to use react table

DorisWongNHS:
TIS21-6543
TIS21-7005


components/forms/ltft/LtftSummary.tsx
```diff
@@ -45,43 +46,51 @@ const LtftSummary = ({
...
-      title="Status date"
+      title="Status"
       data-cy={`table-column_${column.id}`}
     />
   );
 
   //Header value
-  const renderValue = props => <span>{props.renderValue()}</span>;
-  const renderDayValue = props => (
+  const renderValue = (props: any) => <span>{props.renderValue()}</span>;
```
john12321, line 87: We might be able to avoid using the `any` Type here (always a last resort! 😅).
e.g. `props: { renderValue: () => ReactNode }` should work.

components/forms/ltft/LtftSummary.tsx
```diff
@@ -45,43 +46,51 @@ const LtftSummary = ({
...
+  const renderDayValue = (props: any) => (
     <span>{dayjs(props.renderValue()).toString()}</span>
   );
 
   // Operation Column
-  const renderOperationColumnValue = props => (
+  const renderOperationColumnValue = (props: any) => (
```
john12321, line 93: and this should work too:
` const renderOperationColumnValue = (props: {
    row: { original: LtftSummaryObj };
  })`...

john12321 (approved):
LGTM

TODO: need to account for summary table mobile view (e.g. see Cct table for one idea) - but no rush for this merge commit.

PR #1147, opened 2025-02-21
Add designatedBodyCode to PM and include ChangeId in CCT and LTFT DTOs

john12321:
TIS21-7017
TIS21-7018

### DIFF
--- a/components/forms/cct/CctCalcCreate.tsx
+++ b/components/forms/cct/CctCalcCreate.tsx
@@ -192,6 +192,11 @@ export function CctCalcCreate() {
                                     selectedProgramme.endDate,
                                     false
                                   );
+                                  setFieldValue(
+                                    "programmeMembership.designatedBodyCode",
+                                    selectedProgramme.designatedBodyCode,
+                                    false
+                                  );
                                 }
                               } else {
                                 resetForm({ values: defaultCctCalc });
--- a/models/ProgrammeMembership.ts
+++ b/models/ProgrammeMembership.ts
@@ -14,7 +14,6 @@ export interface ProgrammeMembership extends IDateBoxed {
   programmeCompletionDate?: Date;
   curricula: Curriculum[];
   conditionsOfJoining: ConditionsOfJoining;
-  designatedBody?: string;
   designatedBodyCode?: string;
 }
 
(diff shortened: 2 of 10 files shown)


Judge40 (approved):

models/ProgrammeMembership.ts
```diff
@@ -14,6 +14,8 @@ export interface ProgrammeMembership extends IDateBoxed {
   programmeCompletionDate?: Date;
   curricula: Curriculum[];
   conditionsOfJoining: ConditionsOfJoining;
+  designatedBody?: string;
```
ReubenRobertsHEE, line 17: Hey hey now, where does this come from, slipping in the side door so quietly? 🔬

john12321: This whole piece of work is a shoehorning exercise🙈 In my defence, Your Honour, when I added this particular 💩 , I was assuming `designatedBody` would be needed not just DBC. We can always remove it if viewers find this too upsetting? (or pick it up in the next PR?)

components/forms/cct/CctCalcCreate.tsx
```diff
@@ -192,6 +192,11 @@ export function CctCalcCreate() {
                                     selectedProgramme.endDate,
                                     false
                                   );
+                                  setFieldValue(
+                                    "programmeMembership.designatedBodyCode",
+                                    selectedProgramme.designatedBodyCode,
+                                    false
+                                  );
```
DorisWongNHS, line 199: @ReubenRobertsHEE 
I was trying to keep things together, to populate DBC data while we are populating other PM data like programme name, start day and end day, so it is saved in the DB when the POST/PUT api call behind (but not show in the UI). Otherwise, we can add it in the BE to find the matched DBC according to the program when the CCT is create/updated.
Please advise if you have any concern about it.

ReubenRobertsHEE: I was  just wondering about designatedBody, as opposed to designatedBodyCode. I might be missing something, but it only seems to be mentioned once 😁 .

ReubenRobertsHEE:
Thank you, I am much less upset now.

ReubenRobertsHEE (approved):

PR #1151, opened 2025-03-03
Feat/presubmit ltft

DorisWongNHS:
TIS21-6978
TIS21-7055

### DIFF
--- a/components/forms/cct/CctNameModal.tsx
+++ b/components/forms/cct/CctNameModal.tsx
@@ -43,7 +43,7 @@ export function CctNameModal({
               />
               <Button
                 type="submit"
-                disabled={!values.name || isSubmitting}
+                disabled={!values.name.trim() || isSubmitting}
                 data-cy="cct-modal-save-btn"
               >
                 {isSubmitting ? "Saving..." : "Save calculation"}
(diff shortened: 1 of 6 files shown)


ReubenRobertsHEE:
Hello Doris. This looks fine - I was just wondering if we want to allow trainees to name their forms things like " " (space character)? Its not really indicated on the A/Cs, but names that don't have any visible printable characters might cause trouble (for the trainee's, anyway! 😁 )

ReubenRobertsHEE:
I see if you give a CCT calculation a silly name like " " it accepts it, but then fails to save the calculation. So maybe its a slightly wider problem that needs fixing there too - not sure if you agree?

ReubenRobertsHEE (changes requested):
Can I suggest either fixing the blank-string issue for the LTFT name, or writing up a ticket so we don't forget to tackle it later? Otherwise all good 😄 ✅

DorisWongNHS:
@ReubenRobertsHEE thank you for the comments. It's a good shout.
I have added `.trim()` to the LTFT and CCT names, so that the submit button will keep disabled if the name is empty or a space.
The tests are also updated to cover this.
I will add a note to John in case he wants to refactor to validation the fields in other parts of the forms.

ReubenRobertsHEE (approved):

PR #1180, opened 2025-04-08
feat: add Ltft reasons text

john12321:
Add a mix of Warning callout and field hints where appropriate. 
Some of the important text/ next steps still feel a bit vague (e.g. "This will need to be discussed with your regional office."), so may need a bit more tweaking after PM review.

[TIS21-7164](https://hee-tis.atlassian.net/browse/TIS21-7164)


Judge40 (approved):
I think there is definitely some space to improve the messaging, but lets check during review.

PR #1181, opened 2025-04-09
feat: add unsubmitted reason and modified by Role to LTFT Summary table

DorisWongNHS:
New Column to show unsubmit reason in table
TIS21-6706
TIS21-7143

Show application is unsubmitted by LO in status column
TIS21-6706
TIS21-7144


john12321 (approved):
LGTM @DorisWongNHS 
We can tidy up the tooltip placement, text etc. when we have finished with the main implementation.

PR #1190, opened 2025-04-22
Feature Flag - Show LTFT to pilot regions (and beta tester club members)

john12321:
Refactor auth data fetch:
- rename `getCognitoGroups` thunk to `fetchUserAuthInfo` and use this to fetch the `user.features` data.
- Create a new hook `useIsLtftPilot` to check if the user is a LTFT pilot member.
- Delete the old `useIsBetaTester` hook (as the logic is now part of the feature flag BE logic) and all references to it and reference the new `useIsLtftPilot` hook instead to conditionally render the LTFT stuff.

[TIS21-7200](https://hee-tis.atlassian.net/browse/TIS21-7200)

[TIS21-7200]: https://hee-tis.atlassian.net/browse/TIS21-7200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/forms/cct/CctHome.tsx
+++ b/components/forms/cct/CctHome.tsx
@@ -9,16 +9,16 @@ import { useEffect } from "react";
 import { loadCctList } from "../../../redux/slices/cctListSlice";
 import { CctProgrammesList } from "./CctProgrammesList";
 import { fetchLtftSummaryList } from "../../../redux/slices/ltftSummaryListSlice";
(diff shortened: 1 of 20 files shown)


Judge40 (approved):
Approving as the changes are good enough, with a small caveat that naming it a more generic `isLtftEnabled` rather than `isLtftPilot` might be better long term when we start rolling out post-pilot

PR #1196, opened 2025-04-30
fix: user error state

john12321:
Move the user auth actions from the useCriticalDataLoader back to Main so a user failed state (from any source e.g. totp token error) does not stop the app from being displayed.

JFDI


Judge40 (approved):
Struggling with local, but I'll run through it properly on preprod

PR #1200, opened 2025-05-21
chore: update GA scripts

john12321:
... for Hotjar integration.
(GA tag instructions are on https://analytics.google.com/analytics/web/#/a40570867p373187423/admin/streams/table/5142383092 )

chore: Add `gtag` event to `LtftFormView.tsx handleModalFormSubmit` function. This should allow us to first capture the submission event in GA. We should then be able to use this GA event in Hotjar to create a Hotjar event to trigger a survey.

Note: No GA admin access (yet) so can't create a preprod GA data stream to test stuff on.

TIS21-7281

### DIFF
--- a/components/forms/ltft/LtftFormView.tsx
+++ b/components/forms/ltft/LtftFormView.tsx
@@ -100,6 +100,10 @@ export const LtftFormView = () => {
     stopSubmitting();
     setShowModal(false);
     resetAction();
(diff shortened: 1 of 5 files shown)


Judge40 (approved):
Seems simple enough, I'll assume it's pretty standard stuff.

PR #1205, opened 2025-06-03
refactor: remove mentions of RITA from form r

EdwardBarclay:
TICKET TIS21-7267


john12321:
@EdwardBarclay Main is on version 0.126.0 so you'll need to bump to 0.126.1 at least.
Also, no reason I can see why the last e2e test run failed so we'll merge this code and keep our fingers crossed.

PR #1210, opened 2025-06-10
fix: isArcp type bug

john12321:
This commit fixes a bug where the isArcp formData value used in the FormView components is stored as a string type instead of a boolean, which leads to incorrect logic for determining what default linked programme to show and what options to choose from in the pre-submit FormLinkerModal.

NO TICKET


Judge40 (approved):
JS for the win?

john12321:
> JS for the win?

Never a dull moment!

PR #1212, opened 2025-06-18
fix: duplicate draft form bug

john12321:
Add isAutosaving state to FormBuilder 'save & exit' button to disable it from the point a user edits the form until autosave complete to prevent a duplicate draft form (both FormR and LTFT).

The isAutosaving state is set in the useFormAutosave hook via setAutosave setter. The isAutosaving state is then shared via the context children prop to all child components including the FormBuilder component. In the useFormAutosave hook, the setAutosave is set to true outside the setTimeout to ensure there is no window between the user stopping editing and the form save process starting (currently after 2s) which would allow the user to click the 'save & exit' button potentially creating a duplicate draft form.

Aside from disabling the 'save & exit' button, the button text is also changed to 'Saving...' while editing/autosaving.

### DIFF
--- a/components/forms/form-builder/FormBuilder.tsx
+++ b/components/forms/form-builder/FormBuilder.tsx
@@ -113,7 +113,8 @@ export default function FormBuilder({
     setIsFormDirty,
     currentPageFields,
     setCurrentPageFields,
-    jsonForm
+    jsonForm,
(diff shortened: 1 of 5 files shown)


DorisWongNHS (approved):
LGTM

PR #1214, opened 2025-06-23
Feat/ltft pdf

DorisWongNHS:
PDF download button added in LTFT form View Page for all LTFT status
image

Button disabled when form ID is absent
image

TIS21-7324
TIS21-7394

### DIFF
--- a/components/forms/ltft/LtftFormView.tsx
+++ b/components/forms/ltft/LtftFormView.tsx
@@ -35,6 +35,7 @@ import { useActionState } from "../../../utilities/hooks/useActionState";
 import ScrollToTop from "../../common/ScrollToTop";
 import { LtftStatusDetails } from "./LtftStatusDetails";
 import { downloadLtftPdf } from "../../../utilities/FileUtilities";
+import InfoTooltip from "../../common/InfoTooltip";
 
 export const LtftFormView = () => {
   const dispatch = useAppDispatch();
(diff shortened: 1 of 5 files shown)


services/FormsService.ts
```diff
@@ -27,6 +27,19 @@ export class FormsService extends ApiService {
     );
   }
 
+  async downloadTraineeLtftPdf(id: String) {
```
john12321: Hi @DorisWongNHS , I think we can improve this function.
1. Best to use `const` not `let` for the `requestConfig` object as it's not being reassigned.
2. If need `string` type, best practice to use `string` primitive - although I don't think string or any other type is needed because there is no request body for a `get` request.

services/FormsService.ts
```diff
@@ -27,6 +27,19 @@ export class FormsService extends ApiService {
...
+    let requestConfig: AxiosRequestConfig<String> = {
+      headers: {
+        Accept: "application/pdf"
+      },
+      responseType: "blob"
+    };
+    return this.axiosInstance.get<String, AxiosResponse<Blob>>(
```
john12321: 3. I think the first parameter of `axiosInstance.get` should be type `Blob` not `string` as this is the expected res data type.

components/forms/ltft/LtftFormView.tsx
```diff
@@ -115,6 +117,24 @@ export const LtftFormView = () => {
...
+          }}
+        >
+          Save a copy as a PDF
+        </Button>
+        {!formData.id ? (
+          <>
+            &nbsp;
+            <InfoTooltip
+              tooltipId={"pdfButtonInfo"}
+              content="The PDF download button would be disabled if the LTFT is unsaved. Please save before downdloading the PDF."
```
john12321, line 134: Just wondering if we can simplify the messaging here?
Do we need the first sentence?
(If we do then the grammar probably needs tweaking.)
Happy to have a chat about this and any other comments. 😄

john12321 (changes requested):
Mostly looks fine, just some minor things to consider @DorisWongNHS 😄

components/forms/ltft/LtftFormView.tsx
DorisWongNHS: It's a good shout. The first sentence is not necessary. 
Is the sentence "Please save before downloading the PDF." looks alright to you? (I also found the typo in the original sentence) Sorry my English always need advice.

john12321: No worries, your English is infinitely better than my Chinese! Yes I think the second sentence looks fine. Very clear.

PR #1215, opened 2025-07-02
feat: ignore non-REVIEW_DATA actions

ReubenRobertsHEE:
Pending a big rewrite to use them on the action summary.

New action types added in: https://github.com/Health-Education-England/tis-trainee-actions/pull/64

TIS21-7472


ReubenRobertsHEE:
It feels wrong that is it a bit under 1000 lines, so please advise if more code is needed 😁

john12321:
The calm before the storm 😜

john12321 (approved):

PR #1222, opened 2025-07-29
feat: show rejected LTFT

DorisWongNHS:
What is included:
- Show REJECTED application with reason in `Previous Application` table on TSS LTFT tab
- Remove `Created` column in summary tables
- Change `Updated` column header to `Last Updated`
- Summary table filter of REJECTED status
- Shown REJECTED application form detail when row click on summary table
- Display date without timestamp in `Last Updated` column, but shown detailed time when mouseover
- Fix logic of `getStatusReasonLabel()` for UNSUBMITTED and WITHDRAWN to show actual value of reason if no matched label instead of "other". (This fix the value mismatch when LTFT is unsubmitted from TIS with reason `Proposed start date` being shown as `other reason`)

image

TIS21-7524
TIS21-7573
TIS21-7575
TIS21-7576
TIS21-7577

### DIFF
--- a/utilities/ltftUtilities.ts
+++ b/utilities/ltftUtilities.ts
@@ -376,18 +376,18 @@ export function getStatusReasonLabel(
   if (status === "UNSUBMITTED" && statusReason) {
     return (
       ACTION_REASONS.UNSUBMIT.find(reason => reason.value === statusReason)
-        ?.label ?? "Other reason"
+        ?.label ?? statusReason
     );
   }
 
   if (status === "WITHDRAWN" && statusReason) {
     return (
       ACTION_REASONS.WITHDRAW.find(reason => reason.value === statusReason)
-        ?.label ?? "Other reason"
+        ?.label ?? statusReason
     );
   }
 
-  return "";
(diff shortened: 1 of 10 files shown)


utilities/ltftUtilities.ts
```diff
@@ -376,18 +376,18 @@ export function getStatusReasonLabel(
...
   if (status === "WITHDRAWN" && statusReason) {
     return (
       ACTION_REASONS.WITHDRAW.find(reason => reason.value === statusReason)
-        ?.label ?? "Other reason"
+        ?.label ?? statusReason
     );
   }
 
-  return "";
+  return statusReason || "";
```
DorisWongNHS, line 390: This is to fix the value mismatch when LTFT is unsubmitted from TIS Admins-UI with reason `Proposed start date` being shown as `other reason`. Will show the actual value of UNSUBMITTED and WITHDRAWN reason instead of "other".

john12321 (approved):
LGTM @DorisWongNHS but will the e2e gods be kind to you?

PR #1245, opened 2025-09-16
fix: include programmeName in FormR submission

ReubenRobertsHEE:
TIS21-6586

### DIFF
--- a/components/forms/__test__/FormView.test.tsx
+++ b/components/forms/__test__/FormView.test.tsx
@@ -1,21 +1,121 @@
-import { render, act } from "@testing-library/react";
+import { act } from "react";
+import { render, screen } from "@testing-library/react";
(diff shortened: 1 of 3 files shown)


components/forms/form-builder/FormView.tsx
```diff
@@ -106,7 +106,8 @@ export const FormView = ({
       isArcp,
       programmeMembershipId,
       localOfficeName,
-      programmeSpecialty: processedFormData.linkedProgramme?.programmeName
+      programmeSpecialty: processedFormData.linkedProgramme?.programmeName,
```
ReubenRobertsHEE, line 109: Ah, now that looks like quite a fishy assignment to me (though I could have saved myself a lot of hassle by just using programmeSpecialty in the PDF, and assumed it was ok!)

PR #1246, opened 2025-09-17
fix: cypress e2e tests cypress react reference

ReubenRobertsHEE:
Speculative. Works locally. From
https://docs.cypress.io/app/references/changelog:

`The cypress/react18 test harness is no longer included in the Cypress binary. Instead, React 18 support is now shipped with cypress/react`

But this was for cypress 14, so maybe out of date. Nevertheless, intention is to fix broken e2e tests that throw errors like

```
ERROR in ./cypress/component/authentication/CreateTotp.cy.tsx:4:0
Module not found: Package path ./react18 is not exported from package /home/runner/work/tis-trainee-ui/tis-trainee-ui/node_modules/cypress (see exports field in /home/runner/work/tis-trainee-ui/tis-trainee-ui/node_modules/cypress/package.json)

```
for every test.

NO-TICKET

### DIFF
--- a/cypress/component/authentication/ChooseMfa.cy.tsx
+++ b/cypress/component/authentication/ChooseMfa.cy.tsx
@@ -1,7 +1,7 @@
 /// <reference types="cypress" />
 /// <reference path="../../support/index.d.ts" />
 
-import { mount } from "cypress/react";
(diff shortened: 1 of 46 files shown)


ReubenRobertsHEE:
No good

PR #1249, opened 2025-09-19
fix: NTN not being displayed even when users have submitted Form Rs

john12321:
**Problem**
NTN not being displayed due to FormR trainee-action logic not being aligned to new API logic for trainee-actions.

**"Fix"**
As a quick fix to let a user see their NTN, delete the FE FormR logic used to conditionally display the NTN. 
Will fix properly in a follow-up commit.

NO TICKET


ReubenRobertsHEE:
FIX: HACK ;)

ReubenRobertsHEE (approved):

PR #1260, opened 2025-09-30
feat: hide disabled features

Judge40:
Hide nav bar links and home page cards based on the feature flags from the user token.

Some home page cards have individual items which may need disabling, such as COJ and CCT on the Programmes card. These should also be hidden if those feature are not enabled.

TIS21-7657
TIS21-7659
TIS21-7661

### DIFF
--- a/cypress/component/home/Home.cy.tsx
+++ b/cypress/component/home/Home.cy.tsx
@@ -12,7 +12,7 @@ import {
   mockUserFeaturesLtftPilot,
   mockUserFeaturesNonSpecialty,
   mockUserFeaturesSpecialty,
-  mockUserFeaturesUnauthenticated
+  mockUserFeaturesNone
 } from "../../../mock-data/trainee-profile";
 
 const baseCards = ["Support", "MFA"];
@@ -35,10 +35,10 @@ const specialtyCards = [
 const ltftPilotCards = [...specialtyCards, "Changing hours (LTFT)"];
 
 describe("Home with MFA set up", () => {
-  describe("Unauthenticated user", () => {
+  describe("When all features disabled", () => {
     beforeEach(() => {
       store.dispatch(updatedPreferredMfa("SMS"));
-      store.dispatch(updatedUserFeatures(mockUserFeaturesUnauthenticated));
+      store.dispatch(updatedUserFeatures(mockUserFeaturesNone));
       mount(
         <Provider store={store}>
           <Router history={history}>
--- a/cypress/component/navigation/TSSHeader.cy.tsx
+++ b/cypress/component/navigation/TSSHeader.cy.tsx
@@ -13,7 +13,7 @@ import {
   mockUserFeaturesLtftPilot,
   mockUserFeaturesNonSpecialty,
   mockUserFeaturesSpecialty,
-  mockUserFeaturesUnauthenticated
+  mockUserFeaturesNone
 } from "../../../mock-data/trainee-profile";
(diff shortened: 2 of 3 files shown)


utilities/FeatureFlagUtilities.ts
```diff
@@ -0,0 +1,13 @@
+import { FeatureFlag, UserFeaturesType } from "../models/FeatureFlags";
+
+/**
```
john12321, line 3: Fancy!
JSDoc comments trailblazer! 😁

components/home/HomeCards.ts
```diff
@@ -0,0 +1,179 @@
+import { UserFeaturesType } from "../../models/FeatureFlags";
+import { HomeCardItemProps, HomeCardProps } from "../../models/HomeCard";
```
john12321, line 2: can delete this import once type assertions have gone

cypress/component/home/Home.cy.tsx
```diff
@@ -8,37 +8,188 @@ import {
...
-    mount(
-      <Provider store={store}>
-        <Router history={history}>
-          <Home />
-        </Router>
-      </Provider>
-    );
+  describe("Unauthenticated user", () => {
+    beforeEach(() => {
+      store.dispatch(updatedPreferredMfa("SMS"));
```
john12321, line 40: I think the current implementation for an unauthenticated user ("NOMFA") won't allow them to navigate to the Home page (they just get the Support and MFA set-up options in the nav), so not sure we need to test for unauthenticated user here - unless we want to change the logic?

john12321 (changes requested):
Nice implementation. 
The only comment/change I see relates to the visibility of the Home page for unauthenticated user and the tests associated with an unauthenticated user ('NOMFA").
Just looking at my local with NOMFA and the 'home' link is a bit pointless atm, just redirects back to MFA set-up so we don't see the Home page/ cards.
We could change the logic so that the Home nav is visible and the 'home' link on MFA set-up works so we can see the 2 cards (Support and MFA set-up) on the Home page?
Other option is to hide 'home' link on MFA set-up page for "NOMFA" and remove the Home tests for unauthenticated user given they will never see the Home page until authenticated?
What do you think?

cypress/component/home/Home.cy.tsx
Judge40: Agree @john12321, it is a bit odd. I had to fudge the test setup a bit by setting the MFA, as you noted.
I could change the "describe" line, as there is a scenario where the backend will disable all features (no profile found). Though I _think_ in the real world that scenario causes the app to not load too 🤷🏻 

Essentially I'm just trying to test the default, no features enabled, state but the app doesn't let you actually get in to that scenario organically.

john12321: Yeah @Judge40 , the `useCriticalDataLoader` will return an error if the Profile data not loaded so you're right, users never get to that 'default' scenario - unless there are any other real world situations that could cause no features enabled? If not then this is one less thing to worry about and would simplify the logic/tests somewhat.

john12321: I agree, the current logic means we can delete those tests so if you're happy then I'm happy 😁.
(Sorry AFK for about an hour but will have a look at any changes/ approve when I get back if that's OK?)

john12321 (approved):
So when you said you’re “happy enough”… :trollface:

PR #1375, opened 2026-04-22
fix: Profile E2E tests

john12321:
Well, fix one, comment out the other for now as can't get it working with test account.

NO TICKET

### DIFF
--- a/redux/slices/traineeProfileSlice.ts
+++ b/redux/slices/traineeProfileSlice.ts
@@ -68,11 +68,13 @@ export const updateGmc = createAsyncThunk(
   }
 );
 
-export const updateEmail = createAsyncThunk<void, string>(
+export const updateEmail = createAsyncThunk(
   "traineeProfile/personalDetails/updateEmail",
   async (email: string) => {
     const traineeProfileService = new TraineeProfileService();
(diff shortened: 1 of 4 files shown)


redux/slices/traineeProfileSlice.ts
```diff
@@ -68,11 +68,13 @@ export const updateGmc = createAsyncThunk(
...
 
-export const updateEmail = createAsyncThunk<void, string>(
+export const updateEmail = createAsyncThunk(
   "traineeProfile/personalDetails/updateEmail",
   async (email: string) => {
     const traineeProfileService = new TraineeProfileService();
-    await traineeProfileService.updateEmail(email);
+    const response: ApiResponse<PersonalDetails> =
+      await traineeProfileService.updateEmail(email);
+    return response.data;
```
ReubenRobertsHEE, line 77: I'm not sure why these bits have changed, but YOLO

ReubenRobertsHEE (approved):

redux/slices/traineeProfileSlice.ts
john12321: That's the spirit!

PR #1383, opened 2026-05-22
refactor: remove cctDate from ltft form

john12321:
refactor: remove cctDate from ltft form
refactor: remove some more redundant cct date calc logic.

Note: kept the LTFT JSON pretty much the same on this pass (i.e. just removed cctDate but kept the rest as is) but could align it with any BE tidy-up that looks to simplify things. 

TIS21-8664


DorisWongNHS (approved):
Thanks John. Running on local looks good to me.

PR #1387, opened 2026-06-23
Fix/has gmc checkbox

DorisWongNHS:
- Add checkboxes `hasGmcNumber` and `hasGdcNumber` in PH Non-Medic FormR A & B
  When the checkbox is checked the child field is shown; 
  When it is unchecked, the child field is hidden and value in child field will be removed
- Update other fields in forms to make it consistence with the "uncheck and empty child field" behaviour

TIS21-8612
TIS21-8650


john12321:
Hi @DorisWongNHS , I've think I've pulled your branch latest changes... Had a quick look but I don't appear to be seeing the consistent behaviour for the FormB Section 4 isWarned / isComplying i.e. when I toggle to hide the isComplying field the value remains and is saved (looks like other similar contingent fields e.g. immigrationStatus / otherImmigrationStatus are behaving like this too).
It might be a "me" issue (think I've overheated!). Maybe you could run through the changes for me?

components/forms/form-builder/form-fields/Checkboxes.tsx
```diff
@@ -11,6 +11,8 @@ type CheckboxesProps = {
   arrayIndex?: number;
   arrayName?: string;
   dtoName?: string;
+  conditional?: React.ReactNode;
+  conditionalFieldName?: string;
```
john12321: Hi @DorisWongNHS, 
We already have the source of truth for the   hasGmcNumber/hasGdcNumber field mapping so not sure why we are using CONDITIONAL_FIELD_MAP. As discussed the other day, we need one source of truth.

Looking at the code in the FormFieldBuilder and Checkboxes components, as a 'quick fix' we should use the conditionalFieldName prop and delete the hardcoded CONDITIONAL_FIELD_MAP . This gives us a single source of truth and makes the comp generic.

But we really need a more comprehensive solution that works across any similar form field not just these checkboxes. 
I would suggest getting rid of the field-clearing logic entirely (line 66-).  and focus on a solution that reuses the showFormField logic which I briefly mentioned the other day. This is currently only used in setFormRDataForSubmit to strip out the hidden fields before submission (see FormBuilderUtilities comp) but we could create a helper function that uses showFormField e.g. we could call it in the FormContext when isFormDirty?

This is getting deep into the FormBuilder logic so happy to help out with this! 😀

john12321 (changes requested):
Hi @DorisWongNHS please see comments.

DorisWongNHS:
Thanks for testing the changes @john12321 . Not sure why tThe changes in Radios.tsx happen to be missing from the commit.  Anyway it has been refactored now and a bit out of dated (but let me know if the commit still looks weird).

I have added the useEffect in the FormContext.tsx, where it makes all the child field consistent to remove value when parent field is unchecked now. Including the immigration status in Form A, isWarned / isComplying and others in Form B.
Please note this will also include all the declaration sections in Form B, those declarations panel and text area will also be emptied if this is what we want.

john12321:
> Thanks for testing the changes @john12321 . Not sure why tThe changes in Radios.tsx happen to be missing from the commit. Anyway it has been refactored now and a bit out of dated (but let me know if the commit still looks weird).
> 
> I have added the useEffect in the FormContext.tsx, where it makes all the child field consistent to remove value when parent field is unchecked now. Including the immigration status in Form A, isWarned / isComplying and others in Form B. Please note this will also include all the declaration sections in Form B, those declarations panel and text area will also be emptied if this is what we want.

Hi @DorisWongNHS given we need to clear the child GMC/GDC field when hidden, then yes it makes sense to do this for all child fields to make things consistent across all forms. 👍

components/forms/form-builder/form-fields/Checkboxes.tsx
john12321, line 1: Can we get rid of conditionalFieldName prop now?

components/forms/form-builder/FormContext.tsx
john12321, line 1: Showing warning messages and clearing hidden values are 2 distinct tasks so it would make more sense to move the clearing logic into its own useEffect rather than add it to the existing warning msg useEffect. Unless you can see any issue in separating this logic?

components/forms/form-builder/FormContext.tsx
```diff
@@ -166,7 +167,21 @@ export const FormProvider: React.FC<FormProviderProps> = ({
...
+    // remove child field value when depending parent field unchecked
+    if (!isFormDirty) return;
+    setFormData(prevFormData => {
+      const cleaned = currentPageFields.reduce((acc, field) => {
+        if (!showFormField(field, acc)) {
+          return { ...acc, [field.name]: null };
+        }
+        return acc;
+      }, prevFormData);
+      return JSON.stringify(cleaned) === JSON.stringify(prevFormData)
```
john12321: I think we could tidy this 'clear hidden' logic up too...
The current reduce spreads a new form object for every hidden field - even the ones already null, then serialises the form twice on every keystroke, so if we use a 'changed' boolean then have a condition like acc[field.name] != null, then this fixes both issues. 

e.g. something like this:
```
setFormData(prevFormData => {
  let changed = false;
  const cleaned = currentPageFields.reduce((acc, field) => {
    if (!showFormField(field, acc) && acc[field.name] != null) {
      changed = true;
      return { ...acc, [field.name]: null };
    }
    return acc;
  }, prevFormData);
  return changed ? cleaned : prevFormData;
});
```
Note: Used != null instead of !== to give false for both null and undefined (if that's what we want!)

john12321 (changes requested):
Sorry to be a pain!

john12321 (approved):

PR #1388, opened 2026-07-01
Refactor/update to nhsuk 10 and nhs react comps 6

pete-webb-NHSE:
Upgrade nhsuk-frontend library (v10) and nhsuk-react-components (v6) to LTS

This work includes
- upgrade TS from v4.9 to 5.9 for compatibility with the typings shipped by nhsuk-react-components v6.
- upgrade tsconfig.json to match TS: moduleResolution changed from node to bundler — the TS 5 resolution mode that mirrors how Next.js resolves packages, which will correctly resolve the exports-map entry points used by nhsuk-frontend v10 and nhsuk-react-components v6. Also, raise target from es5 to es2017; all browsers we support handle ES2017 natively, so no need to transpile down to ES5 (output is smaller and closer to source).
- replacing all deprecated or changed components to use the latest
- changing and replacing to be as close to the current version as possible with sensible and peer-reviewed changes where appropriate
- refactors current components to semantic HTML where trying to force library components to be the same as current involved too much specific CSS targeting

Also (bit of boy-scouting), remove custom Babel compiler config for prod (it only existed to apply the istanbul coverage plugin - think I had some grand ideas about including e2e test stats in code coverage reports!) and use Next JS's built-in (faster and actively maintained) SWC compiler instead.

TIS21-7717
TIS21-7950
TIS21-7951

### DIFF
--- a/components/forms/cct/CctHome.tsx
+++ b/components/forms/cct/CctHome.tsx
@@ -20,7 +20,7 @@ export function CctHome() {
   return (
     <>
       <WarningCallout data-cy="cct-home-warning">
-        <WarningCallout.Heading>
+        <WarningCallout.Heading visuallyHiddenText="">
           {cctReadBeforeProceedingLabel}
         </WarningCallout.Heading>
         <p data-cy="cct-home-warning-text1">{cctHomeWarningMsgs.text1}</p>
--- a/components/forms/cct/CctHome.tsx
+++ b/components/forms/cct/CctHome.tsx
@@ -20,9 +20,9 @@ export function CctHome() {
   return (
     <>
       <WarningCallout data-cy="cct-home-warning">
-        <WarningCallout.Label visuallyHiddenText={false}>
+        <WarningCallout.Heading>
           {cctReadBeforeProceedingLabel}
-        </WarningCallout.Label>
+        </WarningCallout.Heading>
         <p data-cy="cct-home-warning-text1">{cctHomeWarningMsgs.text1}</p>
         <p data-cy="cct-home-warning-text2">{cctHomeWarningMsgs.text2}</p>
         <p data-cy="cct-home-warning-text3">{cctHomeWarningMsgs.text3}</p>
--- a/components/forms/form-builder/form-sections/ImportantText.tsx
+++ b/components/forms/form-builder/form-sections/ImportantText.tsx
@@ -9,7 +9,10 @@ type ImportantText = {
 export const ImportantText = ({ txtName }: ImportantText) => {
   return (
     <WarningCallout>
-      <WarningCallout.Heading data-cy={`WarningCallout-${txtName}-label`}>
+      <WarningCallout.Heading
+        data-cy={`WarningCallout-${txtName}-label`}
+        visuallyHiddenText=""
+      >
         Important
       </WarningCallout.Heading>
       {displayText[txtName]}
--- a/components/forms/form-builder/form-sections/ImportantText.tsx
+++ b/components/forms/form-builder/form-sections/ImportantText.tsx
@@ -9,12 +9,9 @@ type ImportantText = {
 export const ImportantText = ({ txtName }: ImportantText) => {
   return (
     <WarningCallout>
-      <WarningCallout.Label
-        visuallyHiddenText={false}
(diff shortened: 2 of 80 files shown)


components/forms/cct/CctHome.tsx
```diff
@@ -20,9 +20,9 @@ export function CctHome() {
   return (
     <>
       <WarningCallout data-cy="cct-home-warning">
-        <WarningCallout.Label visuallyHiddenText={false}>
+        <WarningCallout.Heading>
```
ReubenRobertsHEE: On the other items like this you used `visuallyHiddenText=""` to replace `visuallyHiddenText={false}`. Not that this wrong necessarily, just flagging it as a possible omission.

components/forms/form-builder/form-sections/ImportantText.tsx
```diff
@@ -9,12 +9,9 @@ type ImportantText = {
 export const ImportantText = ({ txtName }: ImportantText) => {
   return (
     <WarningCallout>
-      <WarningCallout.Label
-        visuallyHiddenText={false}
-        data-cy={`WarningCallout-${txtName}-label`}
-      >
+      <WarningCallout.Heading data-cy={`WarningCallout-${txtName}-label`}>
```
ReubenRobertsHEE: same note as above

ReubenRobertsHEE:
Is there a reason the `.babelrc` file has not been removed? (there's also a babel-plugin-istanbul in the package which might be superfluous if you get rid of the babel config file).

ReubenRobertsHEE:
Apart from the PanelsCreator changes, which I couldn't reason out in my head 😅  , it all looks solid. I assume there will be minor layout / appearance changes, but is the plan to sanity check these on staging?

ReubenRobertsHEE (changes requested):
Just flagging this for the babel config removal; if there's a reason to keep that then just dismiss the review and I'll approve as-is

gtogbes:
> Is there a reason the `.babelrc` file has not been removed? (there's also a babel-plugin-istanbul in the package which might be superfluous if you get rid of the babel config file).

I think the instanbul plugin showed up in pentest report. Somehow it loads the build environment name as part of  js

pete-webb-NHSE:
@ReubenRobertsHEE @gtogbes I tested removing `.babelrc` and the application still starts successfully, the test/coverage still completes successfully, and Next.js no longer disables SWC because of the the Babel config. Based on that, I think it's safe to remove `.babelrc`. I haven't removed `babel-plugin-istanbul` yet, as I haven't verified whether anything still depends on it

pete-webb-NHSE:
Running `test` and `coverage` with `babel-plugin-instanbul` uninstalled ran successfully, so that also appears to be redundant. I'll commit separately. I'll also take a quick look at those visually-hidden-text attributes. 

> Apart from the PanelsCreator changes, which I couldn't reason out in my head 😅 , it all looks solid. I assume there will be minor layout / appearance changes, but is the plan to sanity check these on staging?

A sanity check on staging sounds sensible. I'm happy to do it, though I'm still getting familiar with the existing UI, so I'd probably miss subtle visual regressions not covered by the tests. If someone who's worked with it more recently could also take a quick look on staging, that would give me a bit more confidence

ReubenRobertsHEE:
> Running `test` and `coverage` with `babel-plugin-instanbul` uninstalled ran successfully, so that also appears to be redundant. I'll commit separately. I'll also take a quick look at those visually-hidden-text attributes.
> 
> > Apart from the PanelsCreator changes, which I couldn't reason out in my head 😅 , it all looks solid. I assume there will be minor layout / appearance changes, but is the plan to sanity check these on staging?
> 
> A sanity check on staging sounds sensible. I'm happy to do it, though I'm still getting familiar with the existing UI, so I'd probably miss subtle visual regressions not covered by the tests. If someone who's worked with it more recently could also take a quick look on staging, that would give me a bit more confidence

Cool, that sounds good. It should be fairly easy to check for visual regressions by flipping between staging and prod, though some bits (exploring the 'setting up your MFA' path for example) will require a bit of effort. Happy to help with a second pair of eyes!

ReubenRobertsHEE (approved):

PR #1389, opened 2026-07-03
fix: hide Founcation programs from FormR linker dropdown

DorisWongNHS:
TIS21-8755

### DIFF
--- a/cypress/component/forms/form-linker/FormLinkerModal.cy.tsx
+++ b/cypress/component/forms/form-linker/FormLinkerModal.cy.tsx
@@ -234,17 +234,19 @@ describe("FormLinkerModal", () => {
     cy.get('[data-cy="isArcp0"]').click();
     cy.get('[data-cy="programmeMembershipId"]').should("exist");
 
-    // Foundation programme should be filtered out and never shown as an option
-    cy.get('[data-cy="programmeMembershipId"]').should(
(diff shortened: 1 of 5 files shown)


cypress/component/forms/form-linker/FormLinkerModal.cy.tsx
```diff
@@ -217,4 +218,33 @@ describe("FormLinkerModal", () => {
...
+        qualifications: []
+      })
+    );
+
+    mountComponent();
+    cy.get('[data-cy="isArcp0"]').click();
+    cy.get('[data-cy="programmeMembershipId"]').should("exist");
+
+    // Foundation programme should be filtered out and never shown as an option
+    cy.get('[data-cy="programmeMembershipId"]').should(
```
john12321: Not sure this assertion (and the one below) targets the list of options to see if Foundation Programme is not one of them i.e. think  it's asserting before the options are rendered.
Might need to open the dropdown first then do the assertion? 
(There is a Cypress cmd called clickSelect that targets the .react-select__option if this helps?)

john12321 (changes requested):
Hi @DorisWongNHS looks good, just one comment around the assertions in FormLinkerModal.cy.tsx

john12321 (approved):

PR #1394, opened 2026-07-28
chore(dev): bypass auth for local dev user

Judge40:
NO-TICKET


utilities/localAuth.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+  }
+
+  const rawToken = (await response.json().then(data => data.token)).trim();
+
+  if (!rawToken) {
+    throw new Error("Local auth token response was empty.");
+  }
+
+  if (rawToken.startsWith('"') && rawToken.endsWith('"')) {
```
john12321: Is this just dead code from the rewrite? (JSON.parse should already handle this etc.)

utilities/localAuth.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  if (!response.ok) {
+    throw new Error(
+      `Local auth token request failed with status ${response.status}.`
+    );
+  }
+
+  const rawToken = (await response.json().then(data => data.token)).trim();
+
+  if (!rawToken) {
```
john12321: Looks like this `!rawToken` code never runs i.e. if issue with token e.g. empty/duff key then data.token will be undefined and throw TypeError before this line.
We could destructure the token from the JSON res and then check its Type/content and throw a duff token error etc.
Something like this:
`const {token} = (await response.json()) ?? {} 
if(typeof token !== "string" || !token.trim()) throw new Error("duff/no token")
return token.trim()`

utilities/localAuth.ts
Judge40: This check guards against a malformed token value like

```json
{
  "token": "\"mytokengoeshere\""
}
```

Which is still a valid JSON response, but the token won't parse without removing the quotes.
It's probably fair to say that we should remove it any just expect the response to be valid, else we'd end up with a massive number of guardrails for very little value.

Judge40:
@john12321 other than the `trim()` causing a TypeError these were attempts to add guardrails for genuine issues that _could_ still occur, give the response structure.

That said,  if the response from nginx returns whitespace or double quotes we should probably just fall over instead of trying to handle every possible bad input scenario.

I'll remove the `trim()` and `slice()` and just leave the check for an empty token in response, since that gives us a reasonable error for debugging. Anything else will just fall over when we try to decode the JWT, which is probably fine for local-only dev handling.

john12321 (approved):

PR #1395, opened 2026-07-31
Tis21 8860 linking hygraph and tss

SuJaneLopez:


components/main/GlobalAlert.tsx
```diff
@@ -1,26 +1,35 @@
...
+  const announcements = useAppSelector(
+    state => state.announcements.announcements
+  );
   const { hasOutstandingActions } = useTraineeActions();
   const pathname = useLocation().pathname;
-  const [recruitDismissed, setRecruitDismissed] = useState(false);
+  const { dismissedIds, dismiss } = useDismissedAnnouncements();
+
+  useEffect(() => {
+    dispatch(fetchAnnouncements());
```
john12321: Is it worth adding an `if (announcementsStatus === "idle" && preferredMfa !== "NOMFA")` above the dispatch to stop component remounts refetching?

SuJaneLopez: Good catch, I've pushed a fix. Thanks!!

john12321 (approved):

PR #1397, opened 2026-08-03
fix: Royal College wording

ReubenRobertsHEE:
NO-TICKET

### DIFF
--- a/utilities/Constants.tsx
+++ b/utilities/Constants.tsx
@@ -384,7 +384,7 @@ export const strDateRegex = /^\d{4}-\d{2}-\d{2}$/;
 // Onboarding Tracker Actions
 const welcomeEmailText =
   "Receive 'Welcome' email 16 weeks prior to starting in post";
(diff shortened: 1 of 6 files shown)


utilities/Constants.tsx
```diff
@@ -384,7 +384,7 @@ export const strDateRegex = /^\d{4}-\d{2}-\d{2}$/;
 // Onboarding Tracker Actions
 const welcomeEmailText =
   "Receive 'Welcome' email 16 weeks prior to starting in post";
-const royalSocietyText = "Register with Royal Society/ Faculty";
```
john12321, line 387: Glad you've kept `royalSocietyText` to remember it by 🥹

ReubenRobertsHEE: If I changed **that** it would immediately balloon into a 5000 line PR, as you well know

john12321: Damn! So close!

john12321 (dismissed):

john12321 (approved):

